PR #30, opened 2020-11-17
Jd Header Footer

JohannDascalescu:
- Added Styling for Header and Footer
- Added new pages(Educational Resources, Help, Project Objectives) contained in Header and Footer
- Added FontAwesome Icons


src/app/app.module.ts
```diff
@@ -5,6 +5,7 @@ import { AppComponent } from './app.component';
 import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
 import { PagesModule } from './pages/pages.module';
 import { ComponentsModule } from './components/components.module';
+import {  } from '@fortawesome/free-solid-svg-icons';
```
Fuhji, line 8: should this be import `{ FontAwesomeModule } from '@fortawesome/angular-fontawesome';`

src/app/components/footer/footer.component.html
```diff
@@ -1 +1,19 @@
...
+    </div>
+    <div>
+        <h3>Help</h3>
+        <a [routerLink]="ROUTES.HELP.route()">Help & Documentation</a>
+        <a>Downloads</a>
+    </div>
+    <div>
+        <h3>About</h3>
+        <a>Partners</a>
+        <a>Socila media</a>
```
Fuhji, line 17: Typo - Social Media

Fuhji (changes requested):
Just a minor typo and looks to be a missing import for fortawesome

src/app/app.module.ts
jon571: I used this package a few times, but I don't think you need to actually import it into a component to use it.

An import at the top of the styles.scss file is enough to get the icons accessible using the normal css classes:

```
@import "@fortawesome/fontawesome-free/css/all.min.css";
```

src/app/components/header/header.component.scss
```diff
@@ -0,0 +1,65 @@
+
+
+
+::ng-deep {
```
jon571, line 4: This will affect all menus in the whole app.  If that's what you want it shouldn't be in this component, if it's not you should use :host::ng-deep to apply it only to this component (and children).

src/app/components/footer/footer.component.html
JohannDascalescu: Fixed :) good spot!

src/app/app.module.ts
jon571: This line can be removed now.

jon571: Ah, it's marked as outdated :-)

Fuhji (approved):

PR #31, opened 2020-11-18
Dw implement map

DanielLyleWilliamWarren:
- installed leaflet
- Added basic leaflet map to quick-map.component.ts


.vscode/settings.json
```diff
@@ -1,5 +1,11 @@
 {
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "editor.formatOnSave": true,
-  "editor.formatOnPaste": false
+  "editor.formatOnPaste": false,
+  "[typescript]": {
+    "editor.defaultFormatter": "vscode.typescript-language-features"
```
Fuhji, line 6: how does this affect the eslint/prettier linting/formatting?

Fuhji: https://github.com/prettier/prettier-vscode/issues/388#issuecomment-489067554

suggests this overrides/disables Prettier

Fuhji: quick testing on branch seems to have disabled prettier

Fuhji (changes requested):
all good but would suggest reverting  vscode/settings.json

PR #33, opened 2020-11-19
Home page and SCSS structure 

Fuhji:
- using material grid layout (needs npm install) for simpler layout
- split components styling into individual files
- first draft of the home page


jon571:
Looks good to me :-)

PR #37, opened 2020-11-19
Dw sr quickmaps retructure

DanielLyleWilliamWarren:
- restructure of components and modules in quickmaps area

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,6 +1,5 @@
 {
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "editor.formatOnSave": true,
-  "editor.formatOnPaste": false,
-  "peacock.color": "#68af4a"
+  "editor.formatOnPaste": false
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,6 @@
 {
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "editor.formatOnSave": true,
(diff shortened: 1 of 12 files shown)


.vscode/settings.json
```diff
@@ -1,5 +1,6 @@
 {
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "editor.formatOnSave": true,
-  "editor.formatOnPaste": false
+  "editor.formatOnPaste": false,
+  "peacock.color": "#68af4a"
```
jon571, line 5: No Way!

DanielLyleWilliamWarren: I swear there was never any issues with peacock until I fiddled with the settings earlier

Fuhji: peacock FTW

src/app/pages/quickMaps/pages/components/sideNavContent/sideNavContent.component.ts
```diff
@@ -0,0 +1,12 @@
+import { Component, OnInit } from '@angular/core';
```
jon571, line 1: Can the path of this just be "src/app/pages/quickMaps/components/sideNavContent/sideNavContent.component.ts"?

Does it need to be in "pages/components" rather than just "components"?

jon571: Obvioulsy @JohannDascalescu doesn't think so because he's merged it!

PR #47, opened 2020-11-24
Js routing update

jon571:
Add query params and route guards in quick maps area.

Not complete as only set for country and micronutrient options, but ready to integrate with sidenav before extending to include all params.


bgsandan:
👍

jon571:
Merging this as it's difficult to test in isolation and I need it for the work with Johann.
Andy gave it a thumbs up anyway :-)

PR #64, opened 2020-11-25
Jyou baseline component fix

jyou-bgs:
- Testing css grid to layout containers.
- Styling data tables to fit in to containers.
- Resized chart elements and added tabs for displaying chartjs and Plotly charts with test data.


.vscode/settings.json
```diff
@@ -1,5 +1,8 @@
 {
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "editor.formatOnSave": true,
-  "editor.formatOnPaste": false
+  "editor.formatOnPaste": false,
+  "[typescript]": {
```
Fuhji, line 5: these need removing as they stop Prettier formatting the code

PR #85, opened 2020-11-26
Sr graph rebuild

Sveouu:
- merged with develop
- added dialog modal on chart component, currently pulling data from one chart


src/app/components/dialogs/chartDialog/chartDialog.component.ts
```diff
@@ -0,0 +1,25 @@
+import { Component, Inject, OnInit } from '@angular/core';
+import { MAT_DIALOG_DATA } from '@angular/material/dialog';
+import { DialogData } from '../baseDialogService.abstract';
+
+export interface ChartDialogData {
+  // eslint-disable-next-line @typescript-eslint/ban-types
```
Fuhji, line 6: We should be using Types instead of generic objects

src/app/components/dialogs/chartDialog/chartDialog.component.ts
```diff
@@ -0,0 +1,25 @@
...
+  selector: 'app-chart-dialog',
+  templateUrl: './chartDialog.component.html',
+  styleUrls: ['./chartDialog.component.scss']
+})
+export class ChartDialogComponent implements OnInit {
+  public content = {};
+  constructor(@Inject(MAT_DIALOG_DATA) public dialogData: DialogData<ChartDialogData>) { }
+
+  ngOnInit(): void {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
Fuhji, line 20: Consequence of using a generic object is that we no longer get the benefits of Typescript, plus having to add exceptions everywhere its used

jyou-bgs (approved):

src/app/pages/quickMaps/pages/baselineDetails/chartCard/chartCard.component.ts
```diff
@@ -54,6 +57,56 @@ export class ChartCardComponent implements OnInit {
     },
   };
 
+  // histo
+  public histograph = {
```
Fuhji, line 61: This needs to be an Interface, and if gets more complex with methods, a class (https://stackoverflow.com/questions/54356711/when-to-use-class-or-interface-in-angular-project)

Fuhji (changes requested):
install @types/chart.js - having Typescript Types will help convert generic objects into Typescript objects

Fuhji (approved):

PR #86, opened 2020-11-26
add api endpoint for sub region data

jon571:
Uses mock data currently, but data looks deficient of some data though, like the micronutrient the data concerns


src/app/apiAndObjects/api/api.service.ts
```diff
@@ -29,6 +30,7 @@ export class ApiService extends BaseApi<DictionaryType> {
   // eslint-disable-next-line @typescript-eslint/member-ordering
   public readonly currentData = {
     getMicronutrientDataOptions: new GetMicronutrientDataOptions(this.USE_LIVE_API),
+    getSubRegionData: new GetSubRegionData(this.USE_LIVE_API),
   };
   // eslint-disable-next-line @typescript-eslint/member-ordering
```
Fuhji, line 35: what does the ESLint rule do thats not acceptable re: ordering?

Fuhji: ah, so its wanting to have `public` followed by `private`.

jon571: Yeah.  I don't want to disable the rule but in this case it's nice to order things in a specific way.

Fuhji: does it impact the logic or is it Personal Preference to not use the Angular standard?https://angular.io/guide/styleguide#member-sequence

jon571: It's more a proximity thing.  It being close to the configure() function is a reminder that if you create a new group it needs to be registered in there.

I'm just having a look at reworking now...

I'm all for the eslint rules, but judging by the number of "// eslint-disable" lines in the codebase at the moment, the current rule set might need some massaging before it's a hassle free process.

Fuhji: Re: proximity - agree it makes sense, although a comment block/header may work as replacement advising others to extend/change

I'd not spotted other disabling (other than svea's PR today), but thats an oversight of the reviewers (thats me missing some too!) not spotting. If there are ESLint rules which are not working we should raise them with rest of devs to decide if its bikeshedding preferences or not.

Fuhji: Made an issue to store these: https://github.com/micronutrientsupport/micronutrient-support-tool/issues/87

jon571: There are a couple of API files that I have to yurn off formatOnSave for otherwise the formatting fails the eslinting. :-(

jon571:
Pushed changes to: 
- add api endpoint for dietary sources
- Also change the way the api.service.ts is structured to minimize eslint disabling

src/app/apiAndObjects/api/api.service.ts
jon571: Changes pushed.  Have another look @Fuhji

jon571:
Build failed due to budgets, probably because of the mocked up data files in the assets.
There's no knock on effects with deployment if I up this to 6mb until the API is up and running and we can clean up assests is there @Fuhji ?

jon571:
add api endpoint for household histogram data

Fuhji:
> Build failed due to budgets, probably because of the mocked up data files in the assets.
> There's no knock on effects with deployment if I up this to 6mb until the API is up and running and we can clean up assests is there @Fuhji ?

No immediate problem if we are serving mock data alongside the project

jon571:
More changes added:
- replace mdns options mock api data with mock data from local file
- increase maximum build size budget to accommodate mock data files

Fuhji:
If the PR is still being worked on you can mark it as Draft (equivalent of GitLab WIP), under the reviewers section

jon571:
Ready for merge now @Fuhji :
- add api endpoint for sub region data
- add api endpoint for dietary sources
- Also change the way the api.service.ts is structured to minimize eslint disabling
- add api endpoint for household histogram data
- replace mdns options mock api data with mock data from local file
- increase maximum build size budget to accommodate mock data files

PR #91, opened 2020-11-27
GitHub action to show how a bundle contents differs from `develop`

Fuhji:
- action to build head(current branch) and base(develop) then generate webpack stats for each
- action to compare the compile a table of diffs for different content within bundle e.g. images
- using a fork of the action within maps org


bgsandan (approved):
Nice 🎖️

Issue #87, opened 2020-11-26
ESLint Rules Discussion


Fuhji:
If you are disabling ESLint rules continually we should review whether they are appropriate. Please add any for discussion to this issue.

Sveouu:
chart.Dialog.compponent.ts 
 --> https://github.com/typescript-eslint/typescript-eslint/blob/v4.7.0/packages/eslint-plugin/docs/rules/ban-types.md
export interface ChartDialogData {
  // eslint-disable-next-line @typescript-eslint/ban-types
  content: object;
}

Fuhji:
> chart.Dialog.compponent.ts
> --> https://github.com/typescript-eslint/typescript-eslint/blob/v4.7.0/packages/eslint-plugin/docs/rules/ban-types.md
> export interface ChartDialogData {
> // eslint-disable-next-line @typescript-eslint/ban-types
> content: object;
> }

We should be using Types not generic objects:

https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/ban-types.md

```
Avoid the Object and {} types, as they mean "any non-nullish value".
This is a point of confusion for many developers, who think it means "any object type".
See this comment for more information.
Avoid the object type, as it is currently hard to use due to not being able to assert that keys exist.
See microsoft/TypeScript#21732.
```

PR #92, opened 2020-11-27
quickmaps navigation updates

jon571:
- quickmaps sub-router implemented
- routerlinks updated to include query params
- sidenav styling tweaked
- e2e tests updated

### DIFF
--- a/src/app/pages/quickMaps/quickMaps.component.ts
+++ b/src/app/pages/quickMaps/quickMaps.component.ts
@@ -1,6 +1,10 @@
 
 import { Component, OnInit } from '@angular/core';
+import { ActivatedRoute, NavigationEnd, Router } from '@angular/router';
+import { QuickMapsRouteData } from './quickMaps-routing.module';
 import { QuickMapsService } from './quickMaps.service';
+import { Subscription } from 'rxjs';
+
 
 @Component({
   selector: 'app-quickmaps',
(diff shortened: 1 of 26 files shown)


jon571:
Passes e2e tests local but fails on server.

May be to do with difference between prod and dev builds.

Will try to fix on Monday :-)

jon571:
Managed to get the e2e tests to fail locally now, so will try to get that fixed...

jon571:
Woop woop, tests pass!

src/app/pages/quickMaps/quickMaps.component.ts
```diff
@@ -0,0 +1,41 @@
...
+  constructor(
+    router: Router,
+    private activatedRoute: ActivatedRoute,
+    public quickMapsService: QuickMapsService,
+  ) {
+    router.events.subscribe((event) => {
+      if (event instanceof NavigationEnd) {
+        let subs: Subscription;
+        // eslint-disable-next-line prefer-const
+        subs = this.activatedRoute.firstChild.data.subscribe((data: QuickMapsRouteData) => {
```
Fuhji, line 27: can this not be const subs and remove the eslint disable?

src/app/pages/quickMaps/quickMaps.service.ts
```diff
@@ -21,6 +21,10 @@ export class QuickMapsService {
   // eslint-disable-next-line @typescript-eslint/member-ordering
   public popGroupIdObs = this.popGroupIdSrc.asObservable();
 
+  private readonly mndDataIdSrc = new BehaviorSubject<string>(null);
+  // eslint-disable-next-line @typescript-eslint/member-ordering
```
Fuhji, line 25: this rule keeps getting pinged. If we are regularly needing to use a private var within a public var, it might be a rule to consider removing https://github.com/micronutrientsupport/micronutrient-support-tool/issues/87

src/app/pages/quickMaps/quickMaps.component.ts
jon571: It didn't like that at runtime.  It was saying that the variable hadn't been initialized.

Fuhji (approved):

bgsandan:
🥇

Issue #146, opened 2021-01-15
White space above map in initial 'Quick MAPS' view


bgsandan:
Map should fill entire panel

jon571:
Fixed in #155

PR #155, opened 2021-01-18
Js quick maps menu

jon571:
quick maps menu updates
- #146 remove white space above map in initial 'Quick MAPS' view
- ensure quick maps menu doesn't show on initial quick maps page (currently does only when first navigating there via app link.)


Fuhji:
I've merged a PR which fixes the e2e tests failing

PR #178, opened 2021-01-18
feat: added full-fat and lightweight version of footer that can be sh…

### LINKED ISSUE #144: Footer not needed within Quick MAPS or other 'tool' pages
Footer from the public info pages takes up too much screen real estate.

Large footer should be removed from these pages entirely.  Scope for a condensed single line for T&Cs, share button etc. but not essential

Fuhji:
- Only show full-fat footer on specific routes which match those defined within `app-routing.module.ts` by passing `showFullFooter: bool`
- Added a lightweight copyright-only footer that can be customised. Will only show when full-fat not needed

closes #144


jon571:
Another way to do this would be to add a data attribute directly to the route, like in the sub-router:
https://github.com/micronutrientsupport/micronutrient-support-tool/blob/develop/src/app/pages/quickMaps/quickMaps-routing.module.ts#L22
The accessing accessing it like here to determine whether to display the "full footer" or not:
https://github.com/micronutrientsupport/micronutrient-support-tool/blob/develop/src/app/pages/quickMaps/quickMaps.component.ts#L30

Fuhji:
> Another way to do this would be to add a data attribute directly to the route, like in the sub-router:
> https://github.com/micronutrientsupport/micronutrient-support-tool/blob/develop/src/app/pages/quickMaps/quickMaps-routing.module.ts#L22
> The accessing accessing it like here to determine whether to display the "full footer" or not:
> https://github.com/micronutrientsupport/micronutrient-support-tool/blob/develop/src/app/pages/quickMaps/quickMaps.component.ts#L30

would you not then need to duplicate this code in every page to determine to show or hide the footer?

jon571:
> > Another way to do this would be to add a data attribute directly to the route, like in the sub-router:
> > https://github.com/micronutrientsupport/micronutrient-support-tool/blob/develop/src/app/pages/quickMaps/quickMaps-routing.module.ts#L22
> > The accessing accessing it like here to determine whether to display the "full footer" or not:
> > https://github.com/micronutrientsupport/micronutrient-support-tool/blob/develop/src/app/pages/quickMaps/quickMaps.component.ts#L30
> 
> would you not then need to duplicate this code in every page to determine to show or hide the footer?

That logic would only be in app.component.ts.

The advantage would be that you wouldn't need to maintain a remote list of pages for it to apply to (minFooterPages), as it's an attribute of the route.

Fuhji:
> > > Another way to do this would be to add a data attribute directly to the route, like in the sub-router:
> > > https://github.com/micronutrientsupport/micronutrient-support-tool/blob/develop/src/app/pages/quickMaps/quickMaps-routing.module.ts#L22
> > > The accessing accessing it like here to determine whether to display the "full footer" or not:
> > > https://github.com/micronutrientsupport/micronutrient-support-tool/blob/develop/src/app/pages/quickMaps/quickMaps.component.ts#L30
> > 
> > 
> > would you not then need to duplicate this code in every page to determine to show or hide the footer?
> 
> That logic would only be in app.component.ts.
> 
> The advantage would be that you wouldn't need to maintain a remote list of pages for it to apply to (minFooterPages), as it's an attribute of the route.

True, but it's also more code for the same end result.

jon571:
> > > > Another way to do this would be to add a data attribute directly to the route, like in the sub-router:
> > > > https://github.com/micronutrientsupport/micronutrient-support-tool/blob/develop/src/app/pages/quickMaps/quickMaps-routing.module.ts#L22
> > > > The accessing accessing it like here to determine whether to display the "full footer" or not:
> > > > https://github.com/micronutrientsupport/micronutrient-support-tool/blob/develop/src/app/pages/quickMaps/quickMaps.component.ts#L30
> > > 
> > > 
> > > would you not then need to duplicate this code in every page to determine to show or hide the footer?
> > 
> > 
> > That logic would only be in app.component.ts.
> > The advantage would be that you wouldn't need to maintain a remote list of pages for it to apply to (minFooterPages), as it's an attribute of the route.
> 
> True, but it's also more code for the same end result.

Technically, it should probably be evaluated on the navigation end event anyway, so that it's only evaluated once per navigation.  If you call a function in an ngIf, it will be triggered on every change detection cycle.  I suppose it's not the most processing hungry function in this case though ;-)

src/app/app.component.ts
```diff
@@ -1,11 +1,34 @@
...
+   */
+  public minFooterPages = ['/quick-maps'];
 
-  constructor() {}
+  constructor(public router: Router, private activatedRoute: ActivatedRoute) {
+    router.events.subscribe((event) => {
+      if (event instanceof NavigationEnd) {
+        let subs: Subscription;
+        // eslint-disable-next-line prefer-const
+        subs = this.activatedRoute.firstChild.data.subscribe((data: any) => {
```
Fuhji, line 25: Using <any> here otherwise was getting inconsistent behaviour with the quickMaps pages. Perhaps interfering with the router subscription within that quickmap component?

Fuhji:
@jon571 ready for review round 2! 🔔

PR #195, opened 2021-01-19
Jd map select

### LINKED ISSUE #143: Selecting a country in 'Baseline' dropdown should naviagte the map to the corresponding country
Similar pan and zoom animation as current selection of a polygon on the map

### LINKED ISSUE #151: Populate country selection list and selection map from API call
https://api.micronutrient.support/country

### LINKED ISSUE #173: Address box shadow/margins/column widths in data tables
- Ensure tables use full space of the card

- Box shadowed table, inside card doesn't look quite right

JohannDascalescu:
- Fixed Table box-shadow, width and spacing issues: closes #173 
- When selecting country from Map or Form they now affect the others selection: closes #143 
- Country selection list and map now populated from API: closes #151


bgsandan (commented):
Visually the output looks good.  I'll leave any comment on structural stuff to @jon571.

Just need to resolve the linting and it should be good to merge :)

src/app/pages/quickMaps/pages/locationSelect/locationSelect.component.ts
```diff
@@ -36,58 +50,143 @@ export class LocationSelectComponent implements OnInit, AfterViewInit {
...
+            fillOpacity: 0.1,
+            color: '#3a1d54',
+            opacity: 0.8,
+          };
+        }
+      });
+    });
+  }
+
+  public getselectedLayer(countryId: string): CountryDictionaryItem {
```
jon571, line 85: I think this function needs a bit of tidying:
- It states it is returning a CountryDictionaryItem but it seems to return a map layer (L.layer?).
- There is no return statement for if this.geojson is falsey (null).  Suggest moving return statement to the very bottom of the function.
- You could add (null === country) to the start of the "if" on line 89 to stop it from still looking for the country after it's already found it, to improve performance, but the saving would be tiny.

src/app/pages/quickMaps/pages/locationSelect/locationSelect.component.ts
```diff
@@ -36,58 +50,143 @@ export class LocationSelectComponent implements OnInit, AfterViewInit {
...
+
+  private resetHighlight(layer: L.Layer): void {
+    if (layer !== this.selectedCountry) {
+      this.geojson.resetStyle(layer);
+    }
   }
 
   // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
-  public highlightFeature(layer: any): void {
+  private hoverHighlightFeature(layer: any): void {
```
jon571, line 117: layer could do with a type.

src/app/pages/quickMaps/pages/locationSelect/locationSelect.component.ts
```diff
@@ -36,58 +50,143 @@ export class LocationSelectComponent implements OnInit, AfterViewInit {
...
+        fillOpacity: 0.3,
+      });
+
+      if (!L.Browser.ie && !L.Browser.edge) {
+        layer.bringToFront();
+      }
+    }
+  }
+
+  private selectFeature(layer: any): void {
```
jon571, line 132: layer could do with a type.

src/app/pages/quickMaps/pages/locationSelect/locationSelect.component.ts
```diff
@@ -17,17 +21,27 @@ export class LocationSelectComponent implements OnInit, AfterViewInit {
 
   public geojson: L.GeoJSON;
   public map: L.Map;
-  public slim: boolean;
 
-  constructor(private http: HttpClient, public quickMapsService: QuickMapsService) {
-    // quickMapsService.sideNavClose$.subscribe((closeSideNavEvent: MouseEvent) => {});
-    quickMapsService.slimObservable.subscribe((slim: boolean) => {
-      this.slim = slim;
+  public selectedCountry;
```
jon571, line 25: Need types for both of these.

src/app/pages/quickMaps/pages/locationSelect/locationSelect.component.ts
```diff
@@ -36,58 +50,143 @@ export class LocationSelectComponent implements OnInit, AfterViewInit {
...
+        onEachFeature: (feature, singleFeatureLayer: L.Layer) => {
+          singleFeatureLayer.on({
+            mouseover: () => {
+              this.hoverHighlightFeature(singleFeatureLayer);
+            },
+            mouseout: () => {
+              this.resetHighlight(singleFeatureLayer);
+            },
+            click: (e) => {
+              this.quickMapsService.setCountryId(`${feature.properties.countryId}`);
```
jon571, line 182: Don't need the `${ or }` here as it's all variable.

src/app/pages/quickMaps/pages/locationSelect/locationSelect.component.ts
```diff
@@ -36,58 +50,143 @@ export class LocationSelectComponent implements OnInit, AfterViewInit {
...
+              this.selectFeature(e.target);
```
jon571, line 183: Do you need this?  It should select via the observable subscription.

src/app/pages/quickMaps/quickMaps.service.ts
```diff
@@ -90,14 +90,13 @@ export class QuickMapsService {
     }
   }
 
-  private updateQueryParams(): void {
+  public updateQueryParams(): void {
```
jon571, line 93: Any reason for changing this? Doesn't look needed.

jon571:
I've been a bit pedantic and added a fair few comments.  I should have started a "review" instead of doing them as individual items, sorry.

src/app/pages/quickMaps/pages/locationSelect/locationSelect.component.ts
JohannDascalescu: Just wanted to try something, forgot to change it back

src/app/pages/quickMaps/pages/locationSelect/locationSelect.component.ts
JohannDascalescu: That's for highlighting the selected country. Is there a way of doing it without it?

jon571: I think it should work without that line as the line above should trigger it.

PR #214, opened 2021-01-22
Js general layout updates

jon571:
Layout style updates to:
- Keep header sticky at the top of the page
- Ensure that when navigating between pages the scrollbar is reset to the top.
- The quickmaps page operates more like an app, with the content area scrollable and the header and sidenav staying still.


jon571:
Hi @Fuhji 
I've encroached on some styling you put in originally for the layout (and wave background) It seems to still work fine for me but can you make sure that all you put in still works?

Fuhji (approved):

bgsandan:
Looks good to me.  I'm going to press the "Big Green Button"

PR #237, opened 2021-01-26
139 map tools layout

### LINKED ISSUE #139: Reduce header size of /maps-tool page to increase visibility of tool option cards
Tool options: Quick MAPS, My MAPS etc. are too far down the page and may be missed 'below the fold'.

Also consider updating to allow 4 cards to sit side-by-side as per the Figma prototype

jon571:
closes #139 
- attempt to reduce top areas of pages enough to show at least a taste of the content below
- switch the main scroll from body to html to interact better when dialogs are shown

### DIFF
--- a/src/app/pages/educationalResources/educationalResources.component.html
+++ b/src/app/pages/educationalResources/educationalResources.component.html
@@ -10,7 +10,7 @@
         elementum enim at nibh egestas. </h2>
     </div>
     <div class="mdc-layout-grid__cell--span-6">
-      <img class="eduCellImage" src="/assets/images/webp/study-map-min.webp">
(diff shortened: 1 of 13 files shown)


jon571:
Hi @Fuhji .
Feel free to say no, no, no this isn't how it is supposed to be done if I've gone against what you originally put in! :-)

Fuhji:
Only comment is the thresholds for responsive causing boxes to now be squashed.

New PR:

image


Original 

image

jon571:
Good spot @Fuhji 
Update pushed.  Better?

Fuhji:
bueno :)

PR #239, opened 2021-01-26
Prepare projections page for future development using gridster grid

Fuhji:
- projection pages using gridster
- moved card component to shared components module for re-use


jon571 (commented):
👍

PR #241, opened 2021-01-26
Dw 175 expand table

DanielLyleWilliamWarren:
- Created new MatTableObject
- chartDialog (may have been renamed by the time you are reading this to expandDialog) now takes a MatTableObject
- Created a table which can be passed any MatTableDataSource object and Array<headers> and will create a table in the expand view.


Fuhji (approved):

src/app/pages/quickMaps/pages/baselineDetails/foodItems/foodItems.component.ts
```diff
@@ -19,51 +10,33 @@ import { CurrentDataService } from 'src/app/services/currentData.service';
...
   selector: 'app-food-items',
   templateUrl: './foodItems.component.html',
   styleUrls: ['./foodItems.component.scss'],
-  changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class FoodItemsComponent implements OnInit, OnDestroy {
+export class FoodItemsComponent implements OnInit {
   @ViewChild(MatPaginator, { static: true }) paginator: MatPaginator;
   @ViewChild(MatSort) sort: MatSort;
-  @Input()
```
jon571, line 35: I'm not sure that you really needed to remove a lot of these lines of code in this file.  I think they're needed.  (Although it doesn't look like it was complete in the first place).

src/app/pages/quickMaps/pages/baselineDetails/monthlyCard/monthlyCard.component.html
```diff
@@ -19,43 +19,43 @@
             <td mat-cell *matCellDef="let element">{{ element.unitPerc }}</td>
           </ng-container> -->
           <ng-container matColumnDef="vegetablesPerc">
-            <th mat-header-cell *matHeaderCellDef mat-sort-header>vegetablesPerc</th>
```
jon571, line 22: It's good to have these.  They don't feed into the dialog though so maybe we could look at passing in column titles to the dialog table in the future too.

jon571 (commented):
Just a couple of pointers.
I think the one about the foofItems component is important to fix, but the other one is additional functionality for the future. :-)

src/app/pages/quickMaps/pages/baselineDetails/foodItems/foodItems.component.ts
Fuhji: true - the ondestroy is needed, the function seems to have been removed:

```
ngOnDestroy(): void {
    this.resizeSub.unsubscribe();
  }
```

src/app/pages/quickMaps/pages/baselineDetails/foodItems/foodItems.component.ts
```diff
@@ -19,51 +10,33 @@ import { CurrentDataService } from 'src/app/services/currentData.service';
...
   constructor(
     private currentDataService: CurrentDataService,
     private quickMapsService: QuickMapsService,
     private dialogService: DialogService,
-    private cdr: ChangeDetectorRef,
-  ) {}
+  ) { }
 
   ngOnInit(): void {
-    this.resizeSub = this.resizeEvent.subscribe((widget) => {
```
Fuhji, line 59: this needs to be re-added

src/app/pages/quickMaps/pages/baselineDetails/foodItems/foodItems.component.ts
```diff
@@ -120,7 +76,6 @@ export class FoodItemsComponent implements OnInit, OnDestroy {
         ],
       },
       options: {
-        maintainAspectRatio: false,
```
Fuhji, line 123: this needs re-adding

src/app/pages/quickMaps/pages/baselineDetails/foodItems/foodItems.component.ts
DanielLyleWilliamWarren: Ah, I had a bit of a conflict resolution issue with that file. Will update accordingly

Fuhji (changes requested):
Think might have pulled from an older master as the foodItems.component is having lots of code removed which is needed

Fuhji (approved):

PR #246, opened 2021-01-26
fix: overflow Text in select #147

### LINKED ISSUE #147: Micronutrient data options' text overflows
Text may be longer than the width of the selection box.  How best to support this?  Text wrapping an option?

JohannDascalescu:
- Removed limit from select width
- Closes #147

### DIFF
--- a/src/app/pages/quickMaps/components/sideNavContent/sideNavContent.component.scss
+++ b/src/app/pages/quickMaps/components/sideNavContent/sideNavContent.component.scss
@@ -158,17 +158,6 @@ h3 {
   }
 }
 
-::ng-deep {
-
-  .mat-select-panel mat-option.mat-option {
-    height: unset;
-  }
-
-  .mat-option-text.mat-option-text {
-    white-space: normal;
-  }
-
-  .mat-tooltip {
-    font-size: 15px;
-  }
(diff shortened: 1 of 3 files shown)


src/app/pages/quickMaps/components/sideNavContent/sideNavContent.component.scss
```diff
@@ -148,9 +150,25 @@ h3 {
...
   }
+
   &.mat-button-disabled {
     & i {
       color: $color_grey;
     }
   }
+}
+
+::ng-deep {
```
jon571, line 161: Don't use ::ng-deep at root or it will affect everything, everywhere.
Use :host ::ng-deep to at least keep it within the component.

Also the sub-selectors need "&" on the start :-)

jon571 (commented):
Just a couple of quick changes needed for the css file. :-)

jon571:
I've re-worked this and just removed the width restriction on the popout list.

The wrapping worked but the unlimited width is better.

I removed the tooltip as it was inconsistent with the other selects.  If a user wants to see the full non-ellipsis version they can click on the select again. :-)

Fuhji:
@bgsandan waiting for your review/approval still?

bgsandan (approved):
Lets hit the big green button™

PR #247, opened 2021-01-27
Added fixed panels to baseline and projection pages

Fuhji:
- added fixed panels showing description and estimate above gridster items
- basic placeholder of content until functionality added in

### Known issues
- fixed panels are within the `<gridster>` object so when dragging the `displayGrid` is visible. You cant move cards on top of the fixed panels but the grid may suggest its possible. Suggestions welcome!

image


jon571:
Only thing that I can suggest about the grid showing is that the those two components could be outside of the gridster area.  They don't need to be in there and could just sit above in the gridster element, in the app-baseline-details template.

All looks good though. :-)

Fuhji:
> Only thing that I can suggest about the grid showing is that the those two components could be outside of the gridster area. They don't need to be in there and could just sit above in the gridster element, in the app-baseline-details template.
> 
> All looks good though. :-)

That was their first location outside the gridster div, but had some CSS issues with gridster scrolling working with the correct height of the div...

Fuhji:
5f026cb removed grid lines from dragging now

PR #260, opened 2021-01-29
Jd sidenav toggle

JohannDascalescu:
- Added open side-nav arrow on minimized view 
- Added slide toggle to lock or unlock the side-nav from minimizing
- Made it so that when you click on view results if not locked side-nav automatically minimizes out of view


src/app/pages/quickMaps/components/sideNavContent/sideNavContent.component.ts
```diff
@@ -42,6 +42,9 @@ export class SideNavContentComponent implements OnInit {
 
   public quickMapsForm: FormGroup;
 
+  public sideNaveToggleLock = new FormControl(false);
+  public lockSideNav = false;
```
jon571, line 46: I don't think that you need both of these.  The formcontrol is probably the right way to go.
I'd suggest that you:
- change the spelling to "sideNavToggleLock" :-)
- Remove the "lockSideNav" variable.
- Remove the "sideNavLockChange" function.
- In the "minimiseSideNav" "if" function check "this.sideNavToggleLock.value" instead.

src/app/pages/quickMaps/components/sideNavContent/sideNavContent.component.scss
```diff
@@ -160,4 +176,42 @@ h3 {
...
   max-width: unset;
+}
+
+mat-slide-toggle {
+  position: absolute;
+  left: 0em;
+  top: 0em;
+}
+
+::ng-deep {
```
jon571, line 187: Using ::ng-deep without nesting isn't good.
You could use a more exact "matTooltipClass" in the template then target the tooltip using "::ng-deep .quickmaps-sidebar-lock-tooltip", for example.

To target the toggle you should put a css class on it ("sidebar-lock-toggle") then target using ".sidebar-lock-toggle ::ng-deep blahhh".

src/app/pages/quickMaps/components/sideNavContent/sideNavContent.component.scss
```diff
@@ -160,4 +176,42 @@ h3 {
...
+  & .mat-tooltip {
+    font-size: 12px;
+  }
+
+  & .mat-slide-toggle-label {
+    padding: 1em;
+  }
+
+  & .mat-slide-toggle-thumb {
+    position: relative !important;
```
jon571, line 198: Don't like all of the "!importants". By giving the toggle a css class and using that to target it (as suggested in the above comment) it shouldn't need them.  If it still does we can work out additional steps, but "!important" should be avoided, apart from when fighting someone else's crap "!important" css :-)

jon571 (changes requested):
After these changes I think we should get the opinion of Andy Bean from a functionality perspective.  I think it works well though! :-)

src/app/pages/quickMaps/components/sideNavContent/sideNavContent.component.scss
```diff
@@ -160,4 +176,42 @@ h3 {
...
+    background: $color_blue_form !important;
+    border-radius: 3px !important;
+    width: 20px !important;
+    height: 15px !important;
+    margin-top: 3px;
+  }
+
+  & .mat-slide-toggle-thumb:before {
```
jon571, line 206: You could also nest this inside the above block and just use:
&:before {

jon571:
Just trying to remember where we were with this one.

I think @JohannDascalescu needs to do a few updates before @bgsandan evaluates :-)

src/app/pages/quickMaps/components/sideNavContent/sideNavContent.component.scss
JohannDascalescu: doesn't seem to like that :/ not sure why it wouldn't work though the more I think about it

jon571 (approved):

jon571:
Assigned to @bgsandan for review :-)

bgsandan:
👍

PR #265, opened 2021-02-01
feat: chart and table now responsive to selected country

DanielLyleWilliamWarren:
- Containers using MAPS scroll style
- Chart and table now responsive to country selected in drop down.


Sveouu:
looks neat!

Issue #301, opened 2021-02-28
Toggle which cards appear in Baseline diet page


bgsandan:
# Summary

Dependent on the data available, not all of the card/charts are valid.  The **baseline** page can either be displaying data as: **"Household level"** or **"country level"**. 

 In **'household' level** mode, the **4 existing cards** should be displayed:

* "Map View"
* "Top 20 Food Items" 
* "Household Dietary Supply"
* "Monthly Food"

For **country level** it should be only the **first 2**:
* "Map View"
*  "Top 20 Food items"

Future cards may appear in one or other mode.

The mode will also impact which API endpoints each card should access

# Details

Exact methods and level of abstraction/encapsulation of 'mode' and config for which cards are visible/cards API endpoints is open for discussion.  

# Wireframes

Suggested wireframes. Not reviewed by Andy yet.

The data level could be displayed on the summary card.

Household:

![household](https://user-images.githubusercontent.com/72566636/109631351-c697f100-7b3d-11eb-8a44-af8cb5e316da.png)

Country:

![country](https://user-images.githubusercontent.com/72566636/109631377-cb5ca500-7b3d-11eb-9241-d299008cb0d3.png)

# Acceptance Criteria

- [x] Diet pages contain a state flag to indicate if in 'household' or 'country' level mode
- [x] Individual cards can be displayed or not displayed dependent on mode
- [x] Household supply and Monthly food cards only appear if in household mode
- [ ] API endpoint used by card for data can differ dependent on which mode the page is displaying
- [x] Mode flag can be set from dietary sources API endpoint (TBC)

Fuhji:
@bgsandan is this extra data level only shown in Baseline page?

Fuhji:
Is there a plan to have the API endpoint in place before the feature freeze @bgsandan ?

bgsandan:
@Fuhji - Yes, only in baseline page makes sense actually

Re: API - Yes, plan is later today and/or pre-stand-up tomorrow :)

jon571:
Have you finished with this area now @Fuhji ?  I've just put this select in on the baseline description component (on my branch).

Fuhji:
The issue is blocked and can't be completed until the API is in place, see https://github.com/micronutrientsupport/micronutrient-support-tool/issues/301#issuecomment-788856288

jon571:
[...]

PR #329, opened 2021-03-03
301 baseline page level updates

jon571:
data level integration for quick maps baseline page
#301 

Currently shows a select if there's more than one option available.  I can remove this if @bgsandan says that it will never ever be the case.


jon571:
Anybody fancy reviewing/merging this?

bgsandan:
At a cursory glance looks ok to me, though won't claim to be as up with the inner workings of the app as @Fuhji / @DanielLyleWilliamWarren

PR #335, opened 2021-03-04
Dw create map menu

DanielLyleWilliamWarren:
- Created settings dialog 'mapViewSettings' for map
- Can now change the colours of the base maps
- stores user chosen colour ramps in localstorage 
- Can be tested [here](https://preview335--micronutrientsupport-tool.netlify.app/quick-maps/baseline?country-id=ATF&mnd-id=2&pop-group-id=1&dataset-id=1&data-level=household)


Fuhji (approved):

src/app/apiAndObjects/objects/enums/colourGradientType.enum.ts
```diff
@@ -0,0 +1,6 @@
+// eslint-disable-next-line no-shadow
+export enum ColourGradientType {
```
jon571, line 2: This enum has really local scope, specifically to the color ramps on those maps, so it would make sense to keep the enum in a more local context, potentially with the actual ramp color codes.

src/app/components/dialogs/dialog.service.ts
```diff
@@ -31,4 +33,8 @@ export class DialogService extends BaseDialogService {
   public openScenarioTypeDialog(): Promise<DialogData> {
     return this.openDialog('scenarioTypeDialog', ScenarioTypeDialogComponent);
   }
+
+  public openMapSettingsDialog(colourGradient: ColourGradientType): Promise<DialogData> {
```
jon571, line 37: Could have some types on the dialogData to make it easier to work out what's being passed in and out.

src/app/components/dialogs/mapSettingsDialog/mapSettingsDialog.component.html
```diff
@@ -0,0 +1,34 @@
...
+        <mat-list-option [selected]="true" value="general">General</mat-list-option>
+        <mat-list-option value="advanced">Advanced</mat-list-option>
+      </mat-selection-list>
+    </mat-drawer>
+
+    <form class="example-sidenav-content" [formGroup]="settingsForm">
+      <div *ngIf="optionList.selectedOptions.selected[0]?.value === 'general'" class="color-box">
+        <p>General</p>
+        <mat-selection-list #colourGradientList formControlName="generalColourSelection"
+          [(ngModel)]="generalSelectionValue" [multiple]="false">
```
jon571, line 14: Using a form control and an ngmodel here seems like overkill.  Better to use just one.

src/app/components/dialogs/mapSettingsDialog/mapSettingsDialog.component.scss
```diff
@@ -0,0 +1,42 @@
+.example-container {
```
jon571, line 1: example?

src/app/components/dialogs/mapSettingsDialog/mapSettingsDialog.component.scss
```diff
@@ -0,0 +1,42 @@
...
+  width: 40vw;
+  height: 50vh;
+  border: 1px solid rgba(0, 0, 0, 0.5);
+}
+
+.example-sidenav-content {
```
jon571, line 7: example?

src/app/components/dialogs/mapSettingsDialog/mapSettingsDialog.component.ts
```diff
@@ -0,0 +1,60 @@
...
+      name: 'Colour Blind',
+    }
+  ];
+
+  public settingsForm = new FormGroup({
+    generalColourSelection: new FormControl('',
+      [Validators.required]),
+  });
+
+  public generalSelectionValue = new Array<string>();
```
jon571, line 42: Using a form control and an ngmodel here seems like overkill.  Better to use just one.

src/app/components/dialogs/mapSettingsDialog/mapSettingsDialog.component.html
```diff
@@ -0,0 +1,34 @@
...
+                [ngClass]="['colour-sample', colourGradientItem.colourGradient]"></span></span>
+          </mat-list-option>
+        </mat-selection-list>
+      </div>
+      <div *ngIf="optionList.selectedOptions.selected[0]?.value === 'advanced'" class="color-box"
+        [ngStyle]="{'background-color': '#2ca25f'}">
+        <p>Advanced</p>
+      </div>
+      <div class="button-wrapper">
+        <button mat-raised-button [disabled]="settingsForm.pristine" (click)="applyChanges()">Apply</button>
```
jon571, line 27: Clicking apply doesn't actually apply, and clicking close does.  Maybe just have one button which is "apply changes" which applies and closes?

src/app/pages/quickMaps/pages/baselineDetails/mapView/mapView.component.ts
```diff
@@ -159,13 +201,82 @@ export class MapViewComponent implements AfterViewInit {
...
+          }</span>` +
+          '</span>';
+      });
+      return div;
+    };
+    this.LegendAbsolute.addTo(this.absoluteMap);
+    this.LegendThreshold.addTo(this.thresholdMap);
+  }
+
   private triggerFitBounds(index: number): void {
```
jon571, line 270: Should probably call this var "tabIndex" to be clear.

src/app/pages/quickMaps/pages/baselineDetails/mapView/mapView.component.ts
```diff
@@ -246,24 +357,24 @@ export class MapViewComponent implements AfterViewInit {
...
-    }).addTo(this.thresholdMap);
+    this.thresholdDataLayer = this.createGeoJsonLayer((feat: GeoJSON.Feature) =>
+      this.getColourRange(null, feat.properties.mnThreshold, this.defaultColourScheme)).addTo(this.thresholdMap);
 
     this.thresholdLegend = new L.Control({ position: 'bottomright' });
 
     this.thresholdLegend.onAdd = () => {
       const div = L.DomUtil.create('div', 'info legend');
-      const range = [0, 20, 40, 60, 80, 99];
+      const range = [0, 10, 20, 40, 60, 80, 99];
```
jon571, line 367: Be nicer if these were closer to the declaration of the colours, maybe in external class(es) too

src/app/pages/quickMaps/pages/baselineDetails/mapView/mapView.component.ts
```diff
@@ -273,47 +384,107 @@ export class MapViewComponent implements AfterViewInit {
...
-        : thresholdValue > 60
-          ? '#f0f01d'
-          : thresholdValue > 40
-            ? '#f08f11'
-            : thresholdValue > 20
-              ? '#e0071c'
-              : thresholdValue > 0
-                ? '#5e1d5c'
-                : '#0f0e0f';
+  private getColourRange(absoluteValue: number, thresholdValue: number, colourGradient: ColourGradientType): string {
```
jon571, line 394: Might be easier to have two parameters here as range (an enum that is "absolute" or "threshold", which is a bit like how you've done the "colourGradient") and "value" for the number value, instead of two variables that only one is populated.

src/app/pages/quickMaps/pages/baselineDetails/mapView/mapView.component.ts
```diff
@@ -273,47 +384,107 @@ export class MapViewComponent implements AfterViewInit {
...
+    if (colourGradient === ColourGradientType.REDYELLOWGREEN) {
+      return absoluteValue > 1500
```
jon571, line 396: It would be better to separate out the threshold and absolute logic here.

This would be better if you used a switch.  I'm not sure if you've seen one like this, but you can arrange it in the following way, instead of the standard way:

```
switch(true) {
  case(thresholdValue > 99): return '#2ca25f';
  case(thresholdValue > 80): return '#addd8e';
  case(thresholdValue > 60): return '#ffeda0';
  case(thresholdValue > 40): return '#feb24c';
  ...
}
```

Or even better, have your array of thresholds and your array of colours then loop:
```
const thresholdColors = [
  {max: 0, color: '#2ca25f'},
  {max: 10, color: '#addd8e'},
  {max: 20, color: '#ffeda0'},
  ...
];
// simplified :-)
function getColour(value): string {
  const thresholdColor = this.thresholdColors.find(thisThresholdColor) => (value < thisThresholdColor.max);
  return (null != thresholdColor ) ? thresholdColor.color : '#123456';
}

```

jon571 (changes requested):
Sorry to have said too much, but what's the point of a review if we don't point out potential improvements?!

src/app/apiAndObjects/objects/enums/colourGradientType.enum.ts
DanielLyleWilliamWarren: It's used in the matSettingsDialog.component.ts, and dialogService too?

src/app/components/dialogs/dialog.service.ts
DanielLyleWilliamWarren: Dialog data is an already defined data Type, you're required to pass in type sensitive data? Do you mean the dataIn and dataOut data types?

src/app/apiAndObjects/objects/enums/colourGradientType.enum.ts
jon571: Ah, I see.  Could still be moved into the map component folder as it only gets passed through the dialog code, not used.

jon571:
I'm happy :-)

jon571 (approved):

PR #336, opened 2021-03-04
Js api updates to live 1

jon571:
Switches to access the live API for the following endpoints on local dev:
- /country
- /micronutrient
- /data-source/{country_id}/{diet/biomarker}

Also makes local dev deployments look at the internal bgs api deployment.

This is also a step in the process of changing the left panel controls so some values are temporarily hard-coded.


package.json
```diff
@@ -13,7 +14,7 @@
     "e2e": "ng e2e",
     "ci:cypress-run": "node --max_old_space_size=8000 ./node_modules/@angular/cli/bin/ng run app:cypress-run",
     "cypress:open": "cypress open",
-    "cypress:run": "cypress run",
+    "cypress:run": "cypress ru",
```
Fuhji, line 17: typo?

jon571: Bugger.... fixing...

PR #354, opened 2021-03-09
fix: tree view working again

Fuhji:
Removing `<div class="bmgf-maps-nice-scrollbar">` from the card fixes the tree card not resizing. Removing from this card fixes immediate issue without impacting other cards which are working okay (AFAICT)


bgsandan (approved):
Approving for now.  @jon571 can complain in a subsequent issue if it broke something :)

Issue #338, opened 2021-03-05
Notification toast for data change


pautva:
# User story
As a nutritionist working with data 
1) I want to see when the app makes an API call that fails and 
2) when the measure of MND changes as a result of changing the selected MN (diet -> biomarker, biomarker -> diet data)

# Description / Example use

We need **snackbars** that inform users of errors and any other changes in the dashboard.

**What happens if a user is on the baseline page and changes the vitamin that's selected to one that is only biomarker, not diet?** 
Switch to that page with the snack bar informing of the change.

More details in the #300 discussion.

# Screenshot of the page

## 1) Error snacbkar (e.g. failed API call)

![some-data-not-loaded](https://user-images.githubusercontent.com/72566636/110305558-b7f58200-7ff4-11eb-831c-11979d986792.png)

## 2) Information snacbkar
![change-measure](https://user-images.githubusercontent.com/72566636/110001717-92265f80-7d0c-11eb-9fe9-19a83ae036a1.png)

# Snackbar examples

## 1) Error snackbar (e.g. failed API call)

![error-snackbar](https://user-images.githubusercontent.com/72566636/110305037-2be35a80-7ff4-11eb-9001-a9331d3818f6.png)

## 2) Information snackbar (e.g. MND change)

![biomarker-snack](https://user-images.githubusercontent.com/72566636/110304544-a495e700-7ff3-11eb-86c8-736affe8317c.png)

![diet-snack](https://user-images.githubusercontent.com/72566636/110304550-a65faa80-7ff3-11eb-87e3-48ec0d5d7962.png)

# Acceptance criteria

- [ ] 1) User is informed with an error snackbar when an API call faills
- [ ] 2) If the user switches the vitamin that's selected to one that is only biomarker, an information snackbar informs the user that different data was loaded.

# Useful links

Snack-bar:
https://material.angular.io/components/snack-bar/overview

Integrate here:
https://github.com/micronutrientsupport/micronutrient-support-tool/blob/develop/src/app/apiAndObjects/api/mapsHttpResponseHandler.ts#L68

Like this:
https://kwvmxgit.ad.nerc.ac.uk/bgs-lims/bgs-lims-gui/-/blob/develop/src/app/apiAndObjects/api/myHttpCallErrorHandler.ts#L33

# Sprint Ready Checklist 
1. - [ ] Acceptance criteria defined 
2. - [ ] Team understands acceptance criteria 
[...]

PR #358, opened 2021-03-10
Dw 338 notifications

DanielLyleWilliamWarren:
- Reusable notifications service and components
- Test buttons currently on header, will remove once approved
- resolves issue #338


jon571 (approved):
Great! :-)

jon571:
It should be noted that this PR doesn't currently resolve the issue #338 though, as it just adds the ability to show notification.  No notifications are actually wired up ;-)

jon571 (approved):
We might want to consolidate all of the "'An error occurred -', 'data could not be loaded'" message s at some point since if that text needs to be updated it'll have to be don in lots of places.  We can do that later.

PR #361, opened 2021-03-10
Added food-group to projected food source

Fuhji:
- projected food source card now has live data wired into the food group / commodity drop down


jon571 (approved):
Great!

PR #365, opened 2021-03-10
feat: update map card api endpoint to use household or country level …

bgsandan:
…data depending on quick maps data mode


jon571 (approved):
Top work :-)

PR #368, opened 2021-03-10
feat: add "Measure of MDN" to quickmaps diet description widgets

jon571:
- add "Measure of MDN" to quickmaps diet description widgets
- Also ensure that the projection version is the same as the baseline one.


Fuhji (approved):
Looks good

Issue #339, opened 2021-03-05
Modal for invalid parameters in Quick Maps URL


pautva:
# User story
As a researcher accessing the MAPS portal from a URL that's been shared with me, I want to be informed if there is something wrong with the URL that I used.

# Description

If a user mistypes the URL that has parameters for search selection, they should be informed that URL is broken and have options to:

* Go back to Quick Maps (if possible try to pre-fill what is valid from the URL params)
* Go back to the homepage

# Page screenshot

![paramateres-error](https://user-images.githubusercontent.com/72566636/110112453-aec4a500-7da9-11eb-908b-fea771aac0e9.png)

# Wireframe 

![erro loading data](https://user-images.githubusercontent.com/72566636/110112527-c69c2900-7da9-11eb-85fa-c2720b31d8f0.png)

# Assets

## Modal message

Invalid parameters were detected in your url. If you are copying and pasting the URL from an email or website, check to see that you haven’t also copied additional punctuation or other text.  

## Image

[Download image](https://drive.google.com/file/d/18aZrXg5TG1ZthhSqKAyufHkVk5FZhYE1/view?usp=sharing)

# Acceptance criteria

- [ ] User sees a notification modal if there is something wrong with the URL they used
- [ ] They can continue with their search afterwards
- [ ] They can go Maps homepage from the modal
- [ ] This is something that can be verified to show that this user story is satisfied.

# Sprint Ready Checklist 
1. - [ ] Acceptance criteria defined 
2. - [ ] Team understands acceptance criteria 
3. - [ ] Team has defined solution / steps to satisfy acceptance criteria 
4. - [ ] Acceptance criteria is verifiable / testable 
5. - [ ] External / 3rd Party dependencies identified

jon571:
The best place to trigger this would be from the routeguard.  We could probably engineer it to show the modal before any redirection but it's easier and probably makes more sense to redirect back to the quickmaps root (location select) while showing the modal.  I've suggested to @JohannDascalescu that this be tried and the content of the modal by updated to:
- state that the redirect has taken place
- only offer the link to the maps homepage since the user will already be on the quickmaps home.

We could replace the "Go To Quick Maps" link with a "Close" link I suppose, to stop a user automatically pressing the link to Maps.

jon571:
Did you see the last comment about the close link @JohannDascalescu ?

PR #373, opened 2021-03-11
Jd error modal

JohannDascalescu:
- created invalid parameter dialog component 
- opens when incorrect url parameters used
- redirects to quick maps page 
- option to be redirected to home page in dialog
issue #339


bgsandan:
@JohannDascalescu are you able to address the merge conflicts in `src/app/pages/quickMaps/quickMapsRouteGuard.service.ts` ?

src/app/components/dialogs/invalidParametersDialog/invalidParametersDialog.component.html
```diff
@@ -0,0 +1,20 @@
+<app-base-dialog [data]="data">
+
+  <div class="container">
+    <h1>The data could not be loaded</h1>
```
jon571, line 4: - h1 might be too big here.  It seems massive!
- The text can be passed into the app-base-dialog anyway, I think at title="The data could not be loaded"

jon571 (approved):

PR #434, opened 2021-04-20
Download charts as png or pdf

Fuhji:


src/app/pages/quickMaps/components/download/download.component.html
```diff
@@ -0,0 +1,23 @@
+<div class="container-row">
+  <div class="element-container">
+    <i class="fas fa-file-image"></i>
+    <a [href]="chartDownloadPNG" target="_blank">
```
bgsandan, line 4: Please could you add the `referrerpolicy="no-referrer-when-downgrade"` attribute?  This will allow the quickchart service to detect the URL query string of the requesting page so the QR code can be updated to point directly to the page the graphic came from?

bgsandan (approved):
Looks great 👍

src/app/pages/quickMaps/components/download/download.component.html
```diff
@@ -0,0 +1,23 @@
...
+      <button *ngIf="chartDownloadPNG" mat-raised-button color="primary">Save as .png</button>
```
jon571, line 5: I think this can be done on the same line using [disabled]="chartDownloadPNG ? true : null".

src/app/pages/quickMaps/components/download/download.component.html
```diff
@@ -0,0 +1,23 @@
...
+      <button *ngIf="!chartDownloadPNG" mat-raised-button color="primary" disabled>Save as .png</button>
+    </a>
+    <p>PNG image of the visualisation that can be used in reports and presentations</p>
+  </div>
+  <div class="element-container">
+    <i class="fas fa-file-image"></i>
+    <a [href]="chartDownloadPDF" target="_blank">
+      <button *ngIf="chartDownloadPDF" mat-raised-button color="primary">Save as .pdf</button>
```
jon571, line 13: Same as above comment.

src/app/pages/quickMaps/components/download/download.component.ts
```diff
@@ -0,0 +1,16 @@
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
```
jon571, line 1: Small file, doesn't look like it should need any of these.

jon571 (commented):
Can't actually test as I'm getting data errors. :-(

src/app/pages/quickMaps/components/download/download.component.html
Fuhji: done

src/app/pages/quickMaps/components/download/download.component.ts
Fuhji: removed

src/app/pages/quickMaps/components/download/download.component.html
Fuhji: added

jon571 (approved):
Looks great!

PR #438, opened 2021-04-21
Updated major dependencies - Will need npm i

Fuhji:
:warning: Will require `npm i`
- Angular 10 --> 11
- Angular Material --> 11
- Gridster to support ng11


jon571 (approved):
You're setting a precedent here. Now were going to have to update every other project to 11! :-)

Issue #419, opened 2021-04-16
Data source drop down in the sidebar for biomarker data


pautva:
# **User story**

As a nutritionist, I want to use the best data source for marker data.

## Summary

When biomarker data is selected and age-gender group is selected, the user should be able to select a data source available. 

If only one data source is available, it should be selected automatically.

If not data source is available, an error message should be displayed.

[Figma prototype](https://www.figma.com/proto/swhv1D7WScZwDE5KKYkkaE/BMGF-Maps?node-id=1767%3A2426&viewport=-4666%2C-4960%2C0.269345760345459&scaling=contain&page-id=0%3A1&hide-ui=1)

## Wireframe

Sidebar
![szoter_annotated_image (2)](https://user-images.githubusercontent.com/72566636/115058305-0b54cd00-9edd-11eb-91af-50156e59a432.jpeg)

No data source available
![image](https://user-images.githubusercontent.com/72566636/115145284-ad44f880-a048-11eb-9342-dee2bba7aa49.png)

# **Related Issues**

## **Front-End**

#416 Add age-gender group dropdown to sidebar when biomarker selected

## **API**

???

## **Database**

???

# **Acceptance Criteria**

*   [ ] When biomarker data and age-gender group is selected, available data sources are loaded in the drop down
*   [x] User can select one of the available data sources
*   [x] If only one source is available, it is selected by default
*   [x] If no source is available, an error message is displayed.

jon571:
Is "Data Source" different from the current "Micronutrient data options"?

pautva:
We have the same dropdown for "Diet Data", but I guess the options in the dropdown for "Biomarker Data" would be different since the data would be coming from a different survey.

jon571:
So we need to change the current "Micronutrient data options" drop-down to "Data Source", and use a different source for the options when "biomarker data" is selected?

bgsandan:
Hi @jon571 Yep, I'm happy for it to use the same dropdown component, routing etc. if that is easier.  The 'Micronutrient Data Options' dropdown could be renamed to something more generic e.g. 'Data Source'.  The key difference will be as @pautva says that the data that populates it will be different.  In the same way that the current diet dropdown calls a different `data-sources` endpoint for `country/household` level diet data, I envisage a third variant of the route for `biomarker` which will return the same structure of data source name and ID which can be used for subsequent requests from the biomarker page cards

jon571:
So an example call for data sources with diet data is:
[...]

PR #439, opened 2021-04-21
419 no data source

jon571:
#419 
If no data source is available, an error message is displayed.

I could only recreate these conditions by altering the underlying data, so it will be difficult to test.

Most of the changes are just renaming variables to dataSource


Sveouu (approved):
Looking good.

PR #450, opened 2021-04-23
416 default age gender group

jon571:
Defaults age-gender selection and only validates it if it is included in query params

THIS IS A PR FOR MERGING INTO THE "sr_416_group_drowpdown" branch


Sveouu (approved):
Nicely laid out to follow the logic  :)

Issue #374, opened 2021-03-11
Integrate share button into quick MAPS pages


bgsandan:
# User story
As a user I want to be able to share the current quick MAPS page I am viewing (including URL query params)

@pautva may have input as to the best place to put this.  In the mini footer perhaps?

# Acceptance criteria

- [x] Share interface from the homepage is also integrated on quick MAPS pages
- [x] Quick MAPS share button shares the current url complete with URL params for query string

pautva:
I think having it at the top of the page would make it easier to find and more likely to use.

I see that we don't have an H1 heading on the dashboard at the moment. It might be useful to include it for accessibility reasons and so people can make sense of where they are quicker. Then the Share button could be next to the heading.

It could look like this:

Diet Data:
![share-button](https://user-images.githubusercontent.com/72566636/110795255-8982db80-826e-11eb-8ce1-7c1e7de20190.png)

Biomarker Data:
![Biomarker Results](https://user-images.githubusercontent.com/72566636/110795269-8e478f80-826e-11eb-8486-44012feefe7c.png)

Fuhji:
Is the Blue-Purple gradient placeholder or Final Form @pautva ?

pautva:
I'd say it's final unless anyone has any objections. I think it makes it stand out slightly, but not too much. It uses one of the purple colours that we already have in UI (#9B51E0) and a new light blue: #00DBDE

CSS properties from Figma:

background: linear-gradient(82.03deg, #00DBDE -1.03%, #9B51E0 100%);
box-shadow: 0px 4px 4px rgba(0, 0, 0, 0.25);
border-radius: 2px;

Fuhji:
Please can you clarify on h1 title requirements as the figma mockup doesn't match the screenshots above @pautva ? Does the Country name need adding to each page title format?

- Baseline: Title?
- Explore Projections to 2050: Title?
- Explore simple dietary change scenarios: Title?
- Explore cost effectiveness scenarios: Title?

pautva:
To keep it consistent with the Biomarker Data page, we should probably just follow the same heading structure for diet data pages: [Nation] Diet Data. 

And I think we can keep the same heading for all the Diet Data tabs.  The tabs themselves serve as a smaller subheading.

**Diet data:**
![image](https://user-images.githubusercontent.com/72566636/116221950-03e7bc00-a746-11eb-9114-1aacf4da902b.png)

**Biomaker data**
[...]

PR #456, opened 2021-04-27
Add share function to quick maps pages

Fuhji:
Affects #374


pautva:
Button works! Should it generate a link to the current Quick Maps view with the query params? At the moment I just get: https://micronutrient.support/

Fuhji:
When i tested in mobile spotted that so fix running through pipeline now

src/app/pages/quickMaps/components/quickMapsHeader.component/quickMapsHeader.component.html
```diff
@@ -1,10 +1,46 @@
+<div class="container">
+  <div class="item-center">
+    <div *ngIf="(quickMapsService.micronutrientObs | async).isDiet">
```
jon571, line 3: It can be diet and biomarker.  Might be better to test the selection using something like this:
quickMapsService.measure === measuretypeEnum.DIET

Fuhji:
@pautva I tested it sharing in Discord......
Working as-expected now

jon571 (changes requested):

pautva:
Link works, but I get a double heading:

image

pautva (approved):
Works as expected now.

src/app/pages/quickMaps/components/quickMapsHeader.component/quickMapsHeader.component.html
Fuhji: How can it be both? The current toggle only allows Diet or Biomarker. What combination allows both?

jon571: image
Iodine allows both so both of the tests you're doing will be true and you get both titles:
https://preview456--micronutrientsupport-tool.netlify.app/quick-maps/diet/baseline?country-id=MWI&mnd-id=I&measure=diet&data-level=household&age-gender-group=all
If you test the which measure has been selected (quickMapsService.measure) rather than what measures are available on the micronutrient you can achieve what you want to :-)

Fuhji: ah ok will make the change cheers

Fuhji: @jon571 this should be fixed now

Fuhji:
@jon571 ready for re-review and merge if happy

Fuhji:
Spotted another thing to change...le sigh

Fuhji:
@jon571 updated cypress and removed the config from the github action abotu chrome/headless, time brought back down to original levels of 12s instead of 20+s

src/app/pages/quickMaps/components/quickMapsHeader.component/quickMapsHeader.component.html
```diff
@@ -1,10 +1,46 @@
...
-  <a class="mode-select-button quickmaps-projection-link" mat-raised-button
-    [routerLink]="((quickMapsService.micronutrientObs | async).isInImpact) ? (ROUTES.QUICK_MAPS_PROJECTION | route) : null"
-    [queryParams]="route.queryParams | async" routerLinkActive="active-route"
-    [disabled]="!(quickMapsService.micronutrientObs | async).isInImpact">Explore projections to 2050</a>
+<div class="container">
+  <div class="item-center">
+    <div *ngIf="quickMapsService.measure === MicronutrientMeasureType.DIET">
+      <h2 class="qmTitle">{{ (quickMapsService.countryObs | async)?.name }} Diet Data</h2>
+    </div>
+    <div *ngIf="quickMapsService.measure !== MicronutrientMeasureType.DIET">
```
jon571, line 6: Because MicronutrientMeasureType is a getter (function) it might mean that it gets called on every change detection cycle.  The easy way to ensure that it isn't would be to have assigned it to a variable in the ts instead:

public readonly MEASURE_TYPE_ENUM = MicronutrientMeasureType;

Having said this, I did exactly the same thing yesterday, so I'm going to test and update my code if what I suspect is true.  If I do need to do that I can update this too if you like, only if I find it is a "problem".

jon571 (approved):

src/app/pages/quickMaps/components/quickMapsHeader.component/quickMapsHeader.component.html
Fuhji: Ok good to flag it up, if its a problem that rears its head we can switcharoony

PR #432, opened 2021-04-20
feat: added MN excess card and updated biomarker component to utilise…

DanielLyleWilliamWarren:
Created MN excess card
Improved scaffolding to allow new cards to be added

PR #458, opened 2021-04-27
feat: simplify biomarker gridster by removing data level driven layou…

jon571:
- Simplify biomarker gridster by removing data level driven layout change, as not needed here, as discussed in this abandoned PR: #432 
- remove possibility of showing data level select from biomarker description component as not applicable, as above
- remove data level display from biomarker description component


jon571:
@bgsandan 
I believe this is the result of what we discussed in the abandoned PR: #432

jon571:
@Fuhji These cypress tests are still intermittently failing due to slow page retrieval.  I'm going to re-run these here :-(

jon571:
Hi @bgsandan 
I'm not sure about this second part of this change now.

In the Diet page we added select to toggle between data levels "Household" and "Country" when they were both available.  The requirement was then shelved, as I remember, but the functionality was left in even though the data never activated it.

This was copied over to the Biomarker version of the component too (although part of the code was missing).  I removed it in this change but am now wondering whether that was right.  The conversion we had suggested that the cards that were displayed wouldn't change depending on the data level (first part of this change), but that doesn't necessarily confirm that the data level will always be the same for the Biomarker page.

What can the data level be on the biomarker page and is it conceivable that a user might be able to change it on this page?

bgsandan:
@jon571 - As far as I am aware this is correct.  For the diet data there were behind the scenes two fundamentally different types of data and methods used depending on the data resolution (household/country).  For biomarkers there is only one type of data as they are by definition measured at the individual person level and there isn't a way of inferring lower resolution data via other proxies/datasets which is essentially what is happening in the country level diet data.  Does that help/make sense?

jon571:
@bgsandan 
YES! That makes sense and better justifies my PR :-)

Although it does raise the question of why the data level is shown on the biomarker page:
image

bgsandan:
Yep, that doesn't need to be there.  I would can it.  Happy to put an issue in for it if that would help?

jon571:
If you could approve this PR then @bgsandan, that would be great :-)

jon571:
> Yep, that doesn't need to be there.  I would can it.  Happy to put an issue in for it if that would help?

Quick little change so added to this PR.

jyou-bgs (approved):

Issue #23, opened 2020-11-13
Progressive Web App


bgsandan:

Fuhji:
this was merged with PR #465 - close this issue for now and re-open a specific one if/when specific service worker functionality is required @bgsandan ?

bgsandan:
Agreed.  Closing

PR #465, opened 2021-04-28
Adding PWA structure

Fuhji:
Issue link #23 

- Added PWA structure 
- Added npm scripts to generate assets and update configuration files as needed. To run, `npm run pwa-gen-assets`
- `ng serve` does not allow you to test service workers, so you have to serve fully. Added an npm script `start-pwa` which builds a prod version and serves using `http-server`.

### ToDo
Add UI to trigger add to home screen functionality (if available) instead of using browser-own UI

### iOS 
iOS Chrome does not allow 'Add to Home Screen' natively, but Safari does:

image

image

### Android

Screenshot_20210428-151810

image

### Lighthouse report

image


bgsandan (approved):
Looks great 👍

jon571 (approved):
Looks good.
Seems to use it on the preview deployment.
Nice one!

PR #472, opened 2021-04-29
build: disable dependabot 

Fuhji:
Deleted `dependabot.yml` file as the PR's are all failing CI pipelines.


bgsandan (approved):
Approving for now, but leaving Issue #473 to ensure we resolve this at the latest in some of the sprint 6 wrap-up tasks

PR #475, opened 2021-04-30
Adds a gridster reset button

Fuhji:
Related issue #469 

- Adds basic reset button to header
- added layout options for baseline and projection
- Refactored gridster layouts within baseline and projection by moving to a service to handle formatting. Should make it easier to add custom layouts and new gridster pages
- ~Not sure about placement of the grid view button (to expand into buttons or dropdown) to be consistent, visible and avoid conflicting with other elements. Had considered adding tiny layout buttons to the footer as an alternative but happy for @pautva to suggest.~ Updated to use floating FAB button with popout options
- Added new accent colour (blue from share button) on @bgsandan suggestion


pautva:
One suggestion could be to fix it at the bottom right. We could display only the icon and expand to icon-and-text on hover.

Then it would always be easily findable in case the layout becomes very messy. Later, it could have some presets for layouts as well.

Reset-layout

bgsandan (approved):
Looks awesome 👍

PR #481, opened 2021-05-06
Dw notification queue

DanielLyleWilliamWarren:
Notifications no longer dismiss but are queued. 

Not necessarily looking to merge, but found it annoying across multiple projects that notifications get dismissed after another is triggered, as there might be important information on display. Might be more appropriate for EPOS @jon571?  Or myHaz manager @Fuhji ? 

@pautva anything feedback/ improvements from a UX perspective?

Test buttons in the header. Error notifications have a longer duration than informative and positive.


jon571:
I think you're right to look for a nice solution to this, but I'm concerned that if there's a queue of notifications waiting to show, the important one that you need to see when it happens won't actually show until the user has stopped looking for it and has moved on to something else.  The notifications get out of sync with what the user is doing.

I personally like the stacking solution to this sort of problem, where the new one is shown at the bottom and the old one is pushed up until it times out.  Am I right in saying that the Angular Material component one doesn't do this?  How did we make that work in EPOS?

 I think you've done the right thing putting it up here to debate! :-)

DanielLyleWilliamWarren:
I think you're right about the user not being able to see an important notification, I was thinking about giving the notifications a priority that allowed some notifications to override and appear straight away. At the moment EPOS is using non-material notifications. Angular have said the snackbar component will never be stackable. Having a queue was the community's solution to not being able to stack.

PR #512, opened 2021-06-03
Further dietary change scaffold

jon571:
add in a dietary change service and connect up the card to changes in the options

Probably gone to far , not having any idea about the data, corresponding api calls, and when they will need to be made, but it shows how some option changes can result in a new API  call and some can just re-use the current data and apply the newly selected value to it.


jon571:
Feel free to merge too if you like as I'll be off tomorrow @Fuhji . :-)

Fuhji (approved):

PR #526, opened 2021-06-09
Dw dietary change scenario maps

DanielLyleWilliamWarren:
- Created map component
- created two maps, match each other's movements using events triggered on move and zoom. 
- passing mock data from parent to map component. 
- data displaying in both maps
- enabled settings menu on card, with access to colour gradient settings.


.vscode/settings.json
```diff
@@ -13,5 +13,28 @@
   },
   "[json]": {
     "editor.defaultFormatter": "esbenp.prettier-vscode"
+  },
+  "peacock.color": "#74ae8f",
```
jon571, line 17: Arrrrrrrrrrrrrrrrrrrrrrrgh!

jon571 (approved):
I think it works well and the code is nice and clean :-)

I think there may need to be a dividing line just to have some separation between the two maps.

I did manage to get the maps out of sync at one point:
image

I do have an idea of how we can fix this :-)

PR #528, opened 2021-06-10
Sr 504 scenarios table

Sveouu:
Scenarios table added based on mock data (SubRegionData)


src/app/pages/quickMaps/pages/dietaryChange/components/comparisonCard/scenarios-table/scenariosTable.component.ts
```diff
@@ -0,0 +1,87 @@
...
+    scenarioDetails.geoJson.features.forEach((item) => {
+      const detail: TableObject = {
+        region: item.properties.subregion_name,
+        scenarioSupply: item.properties.mn_absolute,
+      };
+      scenarioTableObjects.push(detail);
+    });
+
+    const tableData = baselineTableObjects.map((baselineItem: TableObject) => ({
+      ...scenarioTableObjects.find(
```
jon571, line 79: This function looks like it works absolutely fine, but this part is a little in-efficient.

If you were to use [maps](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) for the  "scenarioTableObjects", with a key of the region (name), this last part could be:
```
const tableDatabaselineTableObjects.map((baselineItem: TableObject) => ({
  ...baselineItem,
  (scenarioTableObjectsMap.has(baselineItem.region)) ? ...scenarioTableObjectsMap.get(baselineItem.region) : {},

```

jon571 (approved):
All looks good :-)

If you wanted to improve the efficiency of the "createTableObjects" function, have a go how I suggest :-)

PR #533, opened 2021-07-12
refactor: updated mechanism which allows maps to track each other.

DanielLyleWilliamWarren:
- Using timeout to create a better user experience when tracking maps.

### DIFF
--- a/src/app/pages/quickMaps/pages/dietaryChange/components/comparisonCard/scenariosMap/scenariosMap.component.ts
+++ b/src/app/pages/quickMaps/pages/dietaryChange/components/comparisonCard/scenariosMap/scenariosMap.component.ts
@@ -97,25 +97,19 @@ export class ScenariosMapComponent implements AfterViewInit {
       move: () => {
         this.alignMaps(this.baselineMap, this.scenarioMap);
       },
-      zoom: () => {
-        this.alignMaps(this.baselineMap, this.scenarioMap);
-      },
     });
     this.scenarioMap.on({
       move: () => {
         this.alignMaps(this.scenarioMap, this.baselineMap);
       },
-      zoom: () => {
-        this.alignMaps(this.scenarioMap, this.baselineMap);
-      },
     });
   }
 
-  private alignMaps(interactedMap: L.Map, mirroredMap: L.Map): void {
+  private alignMaps(activatedMap: L.Map, mirroredMap: L.Map): void {
     // wait for inactivity before triggering update
     clearTimeout(this.timeout);
     this.timeout = setTimeout(() => {
-      mirroredMap.setView(interactedMap.getCenter(), interactedMap.getZoom(), { animate: false });
+      mirroredMap.setView(activatedMap.getCenter(), activatedMap.getZoom(), { animate: false });
     }, 50);
   }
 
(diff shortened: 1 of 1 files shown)


src/app/pages/quickMaps/pages/dietaryChange/components/comparisonCard/scenariosMap/scenariosMap.component.ts
```diff
@@ -98,45 +94,23 @@ export class ScenariosMapComponent implements AfterViewInit {
...
-    if (targetMap.getCenter() !== baseMap.getCenter()) {
-      targetMap.panTo(baseMap.getCenter());
-    }
-  }
-
-  private setMapZoom(baseMap: L.Map, targetMap: L.Map): void {
-    if (targetMap.getZoom() !== baseMap.getZoom()) {
-      targetMap.setZoom(baseMap.getZoom());
-    }
+  private alignMaps(baseMap: L.Map, targetMap: L.Map): void {
```
jon571, line 108: You might want to change the name of the variable here as "baseMap" is a bit too similar to "baseline". 
I dunno what to :-)

primary and secondary?
control and mirror?
changing and repeating?

src/app/pages/quickMaps/pages/dietaryChange/components/comparisonCard/scenariosMap/scenariosMap.component.ts
```diff
@@ -98,45 +94,23 @@ export class ScenariosMapComponent implements AfterViewInit {
 
   private initialiseListeners(): void {
```
jon571, line 95: I think you should still have your move and zoom listeners as they were, making it reflect changes straight away.  I just think that after they have called `setMapZoom` or `setMapPosition` they should then call your `alignMaps` function to ensure that everything is aligned after all zooms and movements are done.

DanielLyleWilliamWarren: I had a zoom listener before, something to do with a trigger on animation was causing it so jump around a bit if it wasn't used, that's why the mousedown() listener was being used too. But as I am now using setView() with animation disabled, the move() listener is called when user zooms too, and it is working properly.

jon571: So are you happy with the mirrored map not aligning until the user stops moving then? I thought that was quite nice.

DanielLyleWilliamWarren: It still mirrors and is more responsive than the current implementation as zoom all pan animations are disabled. You can test the current implementation [here](https://develop--micronutrientsupport-tool.netlify.app/quick-maps/diet/dietary-change?country-id=NGA&mnd-id=C&measure=diet&data-level=country&age-gender-group=all)

jon571: I've added @bgsandan as a reviewer to give the stamp of approval.

bgsandan: Looks good to me :+1:

jon571 (approved):

PR #586, opened 2021-09-10
Js api v2 micronutrient availability

jon571:
Replace micronutrient availability endpoint with new version.

Diet -> baseline -> map view will still work, although the data isn't great.

Same with Diet -> scenario -> comparison


Fuhji:
when you say map view works, i only see a map of africa with no data on it? is that what you mean by `the data isn't great.`?

Fuhji (approved):

PR #592, opened 2021-09-10
Stever projection thresholds

Fuhji:
- fix projection thresholds not updating
- fix projection availability chart failing to update after changing the side bar selection
- applied annotation threshold fix to all other charts with threshold line (see random issue with workaround https://github.com/chartjs/chartjs-plugin-annotation/issues/88#issuecomment-374313018)

### DIFF
--- a/src/app/apiAndObjects/objects/misc/chartjsObject.ts
+++ b/src/app/apiAndObjects/objects/misc/chartjsObject.ts
@@ -67,7 +67,7 @@ export interface ChartJSObject {
     annotation?: {
       annotations: Array<{
         type: string;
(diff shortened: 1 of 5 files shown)


Fuhji:
works locally, netlify preview doesnt :-|

Fuhji:
> Seems to work. Couldn't get it to fail in develop though. I am getting a js error, but also getting that in develop, so all good. Probs because of an API change maybe.

go to projections, then see the availability chart = works. Then change the mineral e.g. zinc --> calcium, chart errors out.

Did you clear site data as suggested before testing?

What JS error?

PR #605, opened 2021-10-12
style: mat select override to style scrollbar when selecting

Sveouu:
- applies maps custom scrollbar to angular materials select


jon571:
This doesn't appear to be working for me on the deployment @Sveouu.  Is it working for you?

PR #628, opened 2021-10-14
Monthly food group using new data

### LINKED ISSUE #609: Monthly food groups component needs updating to use new live data structure.
This affects the table, graph and download etc.

Note: The foodGroup dictionary is currently mocked, so the live MonthlyFoodGroup object's foodGroup won't match up and will be null until this dictionary is live.  It's probably better to leave this task until this is live too.

Fuhji:
- close #609 
- fixed issue with top food treeview which was showing wrong colour code for some food groups
- set max value for chart to avoid displaying > 100% in y axis


Fuhji:
@bgsandan @DanielLyleWilliamWarren  ping for review

PR #629, opened 2021-10-15
chore: updated unit & download graphic title

Sveouu:
- updates to Histogram axis & title


Sveouu:
@bgsandan  if cypress fails, what do I need to do ?

PR #658, opened 2021-10-29
Add unmatched totals endpoint

bgsandan:
*   API boilerplate code for new endpoint
*   Update endpoint data when baseline query params change
*   Display outputs in baseline summary card
*   Populate baseline summary "more info" modal

### DIFF
--- a/src/app/apiAndObjects/api/diet/getUnmatchedTotals.ts
+++ b/src/app/apiAndObjects/api/diet/getUnmatchedTotals.ts
@@ -1,0 +1,48 @@
+/* tslint:disable: no-string-literal */
+import { HttpClient } from '@angular/common/http';
+import { DietDataSource } from '../../objects/dietDataSource';
+import { UnmatchedTotals } from '../../objects/unmatchedTotals';
+import { CacheableEndpoint } from '../../_lib_code/api/cacheableEndpoint.abstract';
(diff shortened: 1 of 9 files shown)


src/app/apiAndObjects/api/diet/getUnmatchedTotals.ts
```diff
@@ -0,0 +1,48 @@
+/* tslint:disable: no-string-literal */
+import { HttpClient } from '@angular/common/http';
+import { DietDataSource } from '../../objects/dietDataSource';
+import { UnmatchedTotals } from '../../objects/unmatchedTotals';
+import { CacheableEndpoint } from '../../_lib_code/api/cacheableEndpoint.abstract';
+import { RequestMethod } from '../../_lib_code/api/requestMethod.enum';
+import { MnAvailabilityEndpointHelper } from './mnAvailabilityEndpointHelper';
+
+export class GetUnmatchedTotals extends CacheableEndpoint<
+  Array<UnmatchedTotals>,
```
jon571, line 10: If you only wanted to return one UnmatchedTotals (which it looks like you do when you use it), you can change this from Array<UnmatchedTotals>to UnmatchedTotals (also makes 3rd param redundant), then use the "buildObjectFromResponse" (singular) method rather than "buildObjectsFromResponse".

jon571 (approved):
Added a couple of comments but nothing that would stop merge :-)

PR #659, opened 2021-10-29
chore: add plausible analytics js snippet

bgsandan:


jon571 (approved):
Seems to call for the plausible script and there are no js errors.  Not sure how to test apart from that.

PR #668, opened 2021-11-23
QoL fixes

Fuhji:


src/index.html
```diff
@@ -2,6 +2,8 @@
 <html lang="en">
 
 <head>
+  <script defer data-domain="preview.micronutrient.support"
```
bgsandan, line 5: Is this needed if it is also created in the ngOnInit of the app-component?

Fuhji: oof, yeah that should be there any more. Let me recheck and will tag you when good to go

Fuhji:
the surplus script tag is not visible when use `view source` as before but you can see it is there when ciewing elements in chrome debug tools. The plausible stats page is also showing realtime data to 'follow' you about, so it looks to be still working as before.

bgsandan (approved):

PR #672, opened 2021-11-23
Dw fix expand titles

DanielLyleWilliamWarren:
- Titles appear in expand dialogs accordingly after switching between vitamin or mineral


Fuhji (approved):

Fuhji:
looks all good to me 👍

PR #679, opened 2021-11-24
refactor: simplified the eslint config 

Fuhji:
- using eslint recommended config instead of dozens of custom ones
- tidied up the file exceptions that had built up due to previous eslint config
- fixed all the errors e.g. unused imports or empty methods removed
- slimmed cypress links to custom commands
- changed some `any` to `unknown` types, tests pass but keep an eye out for weirdness
- running eslint shows 28 warnings primarily due to `unexpected any` being used instead of correct types. refactoring would be a slow process so flagging for future reference.


bgsandan (approved):
👏

PR #684, opened 2021-11-25
feat: api choropleth updates

bgsandan:
update api handler to support full response return rather than just data array. 

Add support for setting choropleth bins from API meta response


DanielLyleWilliamWarren (approved):

DanielLyleWilliamWarren:
Tested and looking good!

PR #685, opened 2021-11-25
fix: resolve issue with dialogs using initial accessor values

Fuhji:
- charts, when opened in dialogs, were accessing the wrong quickmap params (getting the initial value, not the latest) which caused issues
- modified the providers/injectables way that dialog service was being initialised to be in sync with the rest of the application state

to test:

- [ ] view baseline chart, open in dialog, change micronutrient, re open chart dialog
- [ ] view projected chart, open in dialog, change micronutrient, re open chart dialog


DanielLyleWilliamWarren:
Nice! Tested and working as expected

DanielLyleWilliamWarren (approved):

Issue #694, opened 2021-11-26
Error when colour gradient is loaded into map on 'Simple Dietary Change Scenario' tab


DanielLyleWilliamWarren:
ERROR TypeError: Cannot read properties of undefined (reading 'length')
    at new ColourGradient (colourGradient.ts:16)
    at ScenariosMapComponent.setColorGradient (scenariosMap.component.ts:277)
    at ScenariosMapComponent.set binRange [as binRange] (scenariosMap.component.ts:46)
    at setInputsForProperty (core.js:10975)
    at elementPropertyInternal (core.js:10019)
    at ɵɵproperty (core.js:14757)
    at ComparisonCardComponent_ng_template_11_Template (comparisonCard.component.html:15)
    at executeTemplate (core.js:9614)
    at refreshView (core.js:9480)
    at refreshEmbeddedViews (core.js:10605)

PR #698, opened 2021-11-26
fix: dietary change scenarios no longer errors when retrieving map colour gradient

DanielLyleWilliamWarren:
resolves issue #694


bgsandan (approved):
Looks good.  Good spot :+1:

PR #703, opened 2021-11-26
Allow missing data options in quickmaps to fail gracefully 

### LINKED ISSUE #689: If pick a country with no data source, the application starts showing broken content
Loading baseline quickmaps page for some countries either does not load, or causes the app to start showing incorrect data:

https://develop--micronutrientsupport-tool.netlify.app/quick-maps?country-id=MAR&mnd-id=Ca&measure=diet&age-gender-group-id=WRA

When selecting a country first, you cannot proceed if no diet data available. However, when changing to this country from another country which has data, the content starts to break down.

Fuhji:
Adds some basic validation of options which contain data and when none available, redirect users to an informative page. This should be re-visited in light of UX as there are many ways to either pre-validate (only show param options which are valid, therefore user can't make a 'mistake') or light-touch UI/UX features like toast indicators to keep user informed about what is happening and why something may not be working/results seem different to expectation.

- if selecting a new location, if there is no data available grey out/disable the view results button
- when viewing a country data, if user changes the side bar parameters, if there is no data available show a Not Found page with some basic information on what to do Currently placeholder text.

closes #689 
### To test:

- [ ] location select, pick an area with data and view results button should be active
- [ ] location select, pick an area with no data and view results button should be disabled
- [ ] viewing valid data in quick maps, select params with no data and should navigate to Not Found page
- [ ] viewing valid data in quick maps, select params with valid data and should navigate to correct data page

### DIFF
--- a/src/app/pages/quickMaps/components/sideNavContent/sideNavContent.component.html
+++ b/src/app/pages/quickMaps/components/sideNavContent/sideNavContent.component.html
@@ -140,5 +140,6 @@
 
 <div class="footer-row" [class.show]="showGoButton"
   [ngClass]="(quickMapsService.slim.obs | async) ? 'slim-sidenav heading' : ''">
-  <button mat-flat-button color="primary" form="quickMapsForm" (click)="submitForm()">View results</button>
+  <button mat-flat-button color="primary" form="quickMapsForm" (click)="submitForm()"
+    [disabled]="btnViewResultsActive === false">View results</button>
 </div>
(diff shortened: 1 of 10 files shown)


bgsandan (approved):
Nice :+1:

pautva:
A bit late to join the discussion, but I think this solution works really good. 

I just noticed that if you **don't select a nation**, but select an MN and then try to select a data source, you get this notification:

image

But if you **select a nation** and MN, you don't get this alert even though there might be no data sources available:
image

---

A few suggestions:

**Scenario 1:** It might be useful to display an error message next to the disabled button to make it a bit more clear why it is disabled:

image

**Scenario 2:** We could be a bit more clear on what dataset was not found and give users an option to go back to the previous selection:

Dataset not available

---

**Assets:**
[Download "No dataset found" SVG illustration](https://drive.google.com/file/d/1DjJFwvlGVL_4ShYCEYhnsF849XV1TqoZ/view?usp=sharing)

Fuhji:
Thanks @pautva will look at implementing your suggestions 👍

PR #748, opened 2022-01-20
fix: get intervention sub routing almost working

bgsandan:
*   Import `InterventionReviewModule` into the QuickMAPS module :star: 
*   Update parents of intervention pages to be the cost-effectivness module rather than intervention\_review
*   Replace some buttons with anchor links to aid in browser preview/debugging of urls
*   Fix review link in compliance page

All intervention pages (except intervention-baseline) work:

e.g. http://localhost:8100/quick-maps/diet/cost-effectiveness/intervention-compliance

http://localhost:8100/quick-maps/diet/cost-effectiveness/intervention-startup-scaleup-costs

Need to sort baseline still :shrug:

### DIFF
--- a/src/app/pages/quickMaps/quickMaps-routing.module.ts
+++ b/src/app/pages/quickMaps/quickMaps-routing.module.ts
@@ -112,10 +112,10 @@ const routes: Routes = [
     loadChildren: () =>
       import('./pages/costEffectiveness/interventionReview.module').then((m) => m.InterventionReviewModule),
   },
-  {
-    path: '**',
-    redirectTo: '',
-  },
+  // {
+  //   path: '**',
+  //   redirectTo: '',
+  // },
 ];
 
 @NgModule({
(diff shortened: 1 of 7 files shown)


src/app/pages/quickMaps/quickMaps.module.ts
```diff
@@ -41,6 +42,7 @@ import { NoResultsComponent } from './pages/noResults/noResults.component';
     PipesModule,
     DietaryChangeModule,
     CostEffectivenessModule,
+    InterventionReviewModule,
```
Fuhji, line 45: this is the only change i hadn't tried!!

src/app/pages/quickMaps/quickMaps-routing.module.ts
```diff
@@ -108,14 +108,14 @@ const routes: Routes = [
...
       import('./pages/costEffectiveness/interventionReview.module').then((m) => m.InterventionReviewModule),
   },
-  {
-    path: '**',
-    redirectTo: '',
-  },
+  // {
+  //   path: '**',
+  //   redirectTo: '',
+  // },
```
bgsandan, line 118: If this wildcard catch is present the above routes for interventions don't seem to work :/

Probably needs fixing so as to not break the QuickMAPS default dropthrough

PR #756, opened 2022-01-26
Dw sr add intervention api

DanielLyleWilliamWarren:
- Created interventionDictionaryItem.ts
- Updated api.service to return interventions from API
- updated intervention creation component to call api and pass avaliable interventions to dialog 
- updated intervention selection dialog to display selectable list of interventions
- user can select interventions, each intervention displayed accordingly

- Not currently persisting, now working on POST call


Fuhji (approved):
👍

PR #764, opened 2022-01-27
Sr dw api flesh continued

DanielLyleWilliamWarren:
- Flesh out all other endpoints of api


src/app/apiAndObjects/api/intervention/getInterventioData.ts
```diff
@@ -0,0 +1,25 @@
+/* tslint:disable: no-string-literal */
```
Fuhji, line 1: we dont use tslint any more - were you getting errors in vscode from the extension which needed this line?

Fuhji (approved):
👍

src/app/apiAndObjects/api/intervention/getInterventioData.ts
DanielLyleWilliamWarren: Removed it in current working branch

PR #765, opened 2022-01-27
Dw 763 scale up costs

DanielLyleWilliamWarren:
This is my impression of what we need to do for all the objects now?


Fuhji:
Yes as I understood the requirements this is the direction we need to head down. Mapping the results of the API calls to strongly typed objects for us to use. Is there an example we can use of taking the API results and mapping to the new object type?

Fuhji (approved):

PR #766, opened 2022-01-27
Dw sr api object finalise

DanielLyleWilliamWarren:
- Added files object and functions for Baseline-assumptions endpoint
- Created objects for all intervention endpoint responses
- updated Intervention data call to use newly created objects
- tested locally


Fuhji (approved):
looks great, only comment is a more general question on how we best maintain consistency with naming object types, but not wanting to bike shed!

PR #797, opened 2022-03-15
feat: Add Unleash Feature Flags

bgsandan:
Adds support for toggling features/routes based on feature flag config.

Highlights:

- New environments for staging-branch, staging-develop and qa alongside production
- featureFlagService for checking feature flag status from typescript modules
- featureFlagDirective utility directive for toggling visibility of HTML components
- featureFlagGuard route guard for disabling routes.

Example:

- Access to the 'Cost Effectiveness' pages restricted by route guard based on feature flag
- 'Explore Cost Effectiveness Scenarios' button disabled status toggled by feature flag

Full write-up and usage guides (to avoid getting lost in MR history) at:

https://github.com/micronutrientsupport/micronutrient-support-tool/wiki/Feature-Flags

### DIFF
--- a/src/app/directives/feature-flag.directive.ts
+++ b/src/app/directives/feature-flag.directive.ts
@@ -9,8 +9,7 @@ export class FeatureFlagDirective implements OnInit {
 
   constructor(
     private viewContainer: ViewContainerRef,
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    private templateRef: TemplateRef<any>,
+    private templateRef: TemplateRef<unknown>,
     private featureFlagsService: FeatureFlagsService,
   ) {}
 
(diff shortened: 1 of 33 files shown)


src/app/directives/feature-flag.directive.ts
```diff
@@ -0,0 +1,26 @@
...
+@Directive({
+  selector: '[appFeatureFlag]',
+})
+export class FeatureFlagDirective implements OnInit {
+  @Input() appFeatureFlag: string | undefined;
+
+  constructor(
+    private viewContainer: ViewContainerRef,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    private templateRef: TemplateRef<any>,
```
DanielLyleWilliamWarren, line 13: I think you can use <unknown> instead of any to avoid having to disable eslint?

DanielLyleWilliamWarren:
> Any idea on why the cypress tests appear to be failing?

It is because the 'share' dialog isn't being triggered properly on click. I investigated yesterday, will fix today. Forgot to bring it up in standup.

Fuhji (approved):
LGTM 👍

PR #802, opened 2022-03-16
Dw branches merge

DanielLyleWilliamWarren:
Scaffold and wireframe of all stages apart from 'Point of consumption' and 'Cost Summary'. All stages are currently read only


src/app/components/dialogs/baseDialogService.abstract.ts
```diff
@@ -54,21 +54,26 @@ export abstract class BaseDialogService {
         panelClass: 'base-dialog',
         data: dialogData,
         id: dialogId,
-        hasBackdrop: (null == parentElement),
-        disableClose: !closable,
+        hasBackdrop: null == parentElement,
+        disableClose: closable,
```
Fuhji, line 58: @DanielLyleWilliamWarren  this was why cypress tests were failing about the share dialog. @Sveouu ive reverted this change to original so if this has unintended consequences (breaks other things!) let me know...

Fuhji (approved):
LGTM 👍

PR #852, opened 2022-06-30
feat: create/load intervention modal

jlczapski:


Fuhji:
cypress tests failing - but ive made quite a few changes to cypress which should resolve them in my PR

src/app/components/dialogs/costEffectivenessSelectionDialog/costEffectivenessSelectionDialog.component.scss
```diff
@@ -1,15 +1,43 @@
+@import '../../../../assets/scss/colour';
```
Fuhji, line 1: can use `@import 'src/assets/scss/_colour.scss';` instead

Fuhji:
image
i dont seem to get a response for ID 2?

Fuhji:
it is possible to enter a blank name currently and submit - would want some basic verification of input

pautva:
> Question for @bgsandan and/or @pautva - how do we stop users copy and editing an existing users copy?

I guess once we have the uuid, it should prevent users from loading the wrong intervention. But if the user loses that uuid, we would want some mechanism for them to retrieve it. 

Another solution could be to create a password for each intervention, and load them only if the passwords match.

bgsandan (commented):
Just a heads up, some of the theming/styling tweaks made for tabs seems to be leaking to other bits of the app:

image

vs the develop branch (correct) version:

image

bgsandan:
Still seeing this behaviour - re-posting in case lost in the Github issue noise.  I assume this isn't intentional?

> Just a heads up, some of the theming/styling tweaks made for tabs seems to be leaking to other bits of the app:
> 
> image
> 
> vs the develop branch (correct) version:
> 
> image

jlczapski:
I think the issue here was that I wasn't calling 'close dialog' method in the promise callback, so it would often be closing the dialog before the data was returned and would be returning null values.. Which is probably also why it would occasionally be successful, as the data _may_ have been loaded before it was closed.

Fuhji (approved):

PR #853, opened 2022-06-30
feat: url stores selected intervention ids and loads them on page load

kerberpolis:
- When a user selects an intervention id its id value is inserted in the url
- When the user loads the cost effectiveness tab it loads the interventions that are stored in the url

### DIFF
--- a/src/app/pages/quickMaps/pages/costEffectiveness/components/interventionCreation/interventionCreation.component.ts
+++ b/src/app/pages/quickMaps/pages/costEffectiveness/components/interventionCreation/interventionCreation.component.ts
@@ -1,4 +1,4 @@
-import { ChangeDetectionStrategy, ChangeDetectorRef, Component } from '@angular/core';
+import { ChangeDetectorRef, Component } from '@angular/core';
 import { DictionaryType } from 'src/app/apiAndObjects/api/dictionaryType.enum';
 import { InterventionsDictionaryItem } from 'src/app/apiAndObjects/objects/dictionaries/interventionDictionaryItem';
 import { Dictionary } from 'src/app/apiAndObjects/_lib_code/objects/dictionary';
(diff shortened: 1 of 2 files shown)


src/app/pages/quickMaps/pages/costEffectiveness/components/interventionCreation/interventionCreation.component.ts
```diff
@@ -1,37 +1,82 @@
...
 
+  ngAfterViewInit() {
+    this.loadInterventions()
+  }
+
+  public async loadInterventions(): Promise<any> {
+    const searchParams = new URLSearchParams(window.location.search);
+    const inteventionIds = searchParams.get('intIds') ? JSON.parse(searchParams.get('intIds')) : [];
+    for (const id of inteventionIds) {
+      await this.interventionService.getIntervention(id).then((data: any) => {
```
Fuhji, line 45: use a named Type where possible, ESLint should flag this up as a warning if not 

in this case `getIntervention` returns `InterventionData`

src/app/pages/quickMaps/pages/costEffectiveness/components/interventionCreation/interventionCreation.component.ts
```diff
@@ -1,37 +1,82 @@
...
     void dictionariesService.getDictionaries([DictionaryType.INTERVENTIONS]).then((dicts: Array<Dictionary>) => {
       this.interventionsDictionaryItems = dicts.shift().getItems();
     });
   }
 
+  ngAfterViewInit() {
+    this.loadInterventions()
+  }
+
+  public async loadInterventions(): Promise<any> {
```
Fuhji, line 41: return a Type, not `any`. If not returning anything return `void` e.g. `Promise<void>`

PR #861, opened 2022-07-01
Intervention page tweaks

Fuhji:
- removed Point of consumption from sidebar stepper
- industry info table source column added, rows check for rowUnits to decide if percentage or number
- monitoring info table source column added, rows check for rowUnits to decide if percentage or number


Fuhji:
@DanielLyleWilliamWarren @bgsandan bumping this MR from last week....(only small)

bgsandan (approved):
LGTM 👍

PR #863, opened 2022-07-04
Dw 848 service id access

DanielLyleWilliamWarren:
- Each view retrieves Id of intervention, 
-  if ID is invalid, route back to intervention creation view.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3412,7 +3412,7 @@
     "@types/json5": {
       "version": "0.0.29",
       "resolved": "https://registry.npmjs.org/@types/json5/-/json5-0.0.29.tgz",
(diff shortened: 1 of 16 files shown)


Fuhji:
I was able to access a page directly in a browser, using the URL copies from another browser that had run through the intervention process:
- select intervention id
- click review
- continue through couple pages
- copy the URL e.g. `https://preview863--micronutrientsupport-tool.netlify.app/quick-maps/diet/cost-effectiveness/intervention-consumption?country-id=ETH&mnd-id=Fe&measure=diet&age-gender-group-id=WRA`
- open another browser (tried Safari and Firefox), paste the URL
- page loads at the point of consumption page
- When try to move through intervention stages forwsrd/backward it kicks out to start page as expected
- Note: i tried this on other intervention pages are it directly boots you out to the start page so looks like i got a random hit
- Note2: point of consumption was removed from a recent MR so this point may be moot!

Fuhji (changes requested):
LGTM just needs develop merging in beforehand

Fuhji (approved):

PR #873, opened 2022-07-08
feat: track user input changes

jlczapski:
Started investigating how we can track updates from individual inputs in the table. 

```html
<form [formGroup]="form">
    // Ex. input
    <input type="number" matInput required [value]=element.year0 formControlName="year0">
</form>
```

```javascript
public form: FormGroup;
public formChanges: Record<string, unknown> = {};

this.form =  new FormGroup({
  year0: new FormControl(0),
  year1: new FormControl(0),
  year2: new FormControl(0),
});

this.form.valueChanges.pipe(
      pairwise(),
      map(([oldState, newState]) => {
        const changes = {};
        for (const key in newState) {
          if (oldState[key] !== newState[key] && oldState[key] !== undefined) {
            changes[key] = newState[key];
          }
        }
        return changes;
      }),
      filter(changes => Object.keys(changes).length !== 0 && !this.form.invalid)
    ).subscribe(value => {
      console.log('Form has changed: ', value);
      this.formChanges[Object.keys(value).shift()] = Object.values(value).shift();
      console.log(this.formChanges);
    });
```

Currently I have just created 3 variables for inputs year0, year1 & year2. We need to consider how we will generate vars for all 40 inputs: 

- Use *ngFor ?
- Create method to initialise all inputs in ngOnInit  

@kerberpolis tagging you to update as I know you're also investigating approaches currently.

### DIFF
--- a/src/app/pages/quickMaps/pages/costEffectiveness/interventionReview/pages/interventionIndustryInformation/interventionIndustryInformation.component.html
+++ b/src/app/pages/quickMaps/pages/costEffectiveness/interventionReview/pages/interventionIndustryInformation/interventionIndustryInformation.component.html
@@ -4,141 +4,143 @@
     <div class="title">
       <span> Industry Information </span>
     </div>
-    <table mat-table [dataSource]="dataSource">
-      <ng-container matColumnDef="labelText">
-        <th class="row-title" mat-header-cell *matHeaderCellDef> Intervention year </th>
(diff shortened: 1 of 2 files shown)


kerberpolis:
@jlczapski I have just expanded on you work done here. Here is my PR https://github.com/micronutrientsupport/micronutrient-support-tool/pull/876

PR #881, opened 2022-07-12
Force only two input decimals to be used in specific inputs

Fuhji:
- forked from @Sveouu branch `origin/sr_recurring_costs_dialog`
- directive was not being called so moved import out of the component and into the parent module
- inputs do not enforce the directive of 2 decimal places when `type="number"` so using `text`
- column totals calculated by only using items with `rowType = US Dollars`. Note: will need to update to allow modifier inputs e.g. multiple total by a percentage
- standardised buttons to be flat (stroked) instead of raised
- rollback some tab ui changes which were breaking other parts of the ui
- fixed intervention dialog buttons to match ui and be less verbose
- PauliusPurple:tm: added to VSCode Peacock

### TO DO
- if the changes are valid, roll out to other inputs needing this (@bgsandan are all inputs desirable to be 2 decimal places?)

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -10,7 +10,7 @@ jobs:
   release-please:
     runs-on: ubuntu-latest
     steps:
-      - uses: GoogleCloudPlatform/release-please-action@v3.1.4
+      - uses: GoogleCloudPlatform/release-please-action@v3.2.8
         with:
           release-type: node
           package-name: release-please-action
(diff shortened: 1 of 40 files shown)


bgsandan:
When switching from `<input type="number">` to text input, if you use `<input inputmode="numeric" />` then it will still get the numeric keyboard e.g. on a mobile device ⚡

Sveouu (commented):
Directive working as intended

jlczapski (approved):

PR #882, opened 2022-07-12
submit industry info changes

kerberpolis:
- add function to intrventiondataservice that calls a PATCH request with changes to industry information
- BehaviourSubject added to interventionDataService to track changes across components. call subject.next({...subject.value, ...this.changes}) in each component when form inputs change to update the subjects value with the latest changes. On submit, get final value to send your PATCH request.


jlczapski:
https://github.com/micronutrientsupport/micronutrient-support-tool/pull/883

PR #890, opened 2022-07-21
Extra work done on MR 883

Fuhji:
Addition to #883 

- each intervention GET is using the live API instead of cached data. This means that when moving between intervention pages it pulls the latest info each time, which includes any changes made in other pages.
- moved the store and send PATCH to intervention service which is called when confirming page, instead of when at cost summary page
- update industry intervention page to use string inputs instead of numbers, to enable 2 decimal place pipe to work
- ...which meant that needed to ensure that inputs were changes to numbers when building the PATCH request

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -14,24 +14,5 @@
   "[json]": {
     "editor.defaultFormatter": "esbenp.prettier-vscode"
   },
-  "workbench.colorCustomizations": {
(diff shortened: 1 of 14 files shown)


Fuhji:
@jlczapski  @kerberpolis This branch adds to your PR but should be working for both industry and monitoring pages now. Please take a look and comment

jlczapski (approved):

PR #892, opened 2022-07-22
More use of PATCH endpoint for interventions

Fuhji:
- startup costs with dialogs patching data changes
- recurring costs with dialogs patching data changes
- updated industry information to track percentage data not just number
- added blue style to inputs which are modified

## How to test
- look in the interventions pages at recurring costs and startup costs. Open the table dialogs, change values (which should show as blue when edited), confirm to close and save the changes, then reopen the dialog/refresh page and data should persist

## Note
- `.toFixed(2)` on input values are not being honoured. The appDecimal pipe ensures that user inputs are 2dp but if the original data is more than 2dp the `toFixed(2)` does not work.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -14,5 +14,24 @@
   "[json]": {
     "editor.defaultFormatter": "esbenp.prettier-vscode"
   },
-  "peacock.color": "#9b51e0"
+  "peacock.color": "#9b51e0",
+  "workbench.colorCustomizations": {
+    "activityBar.activeBackground": "#b47ce8",
+    "activityBar.activeBorder": "#f1d0ae",
+    "activityBar.background": "#b47ce8",
+    "activityBar.foreground": "#15202b",
+    "activityBar.inactiveForeground": "#15202b99",
+    "activityBarBadge.background": "#f1d0ae",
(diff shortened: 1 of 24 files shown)


.vscode/settings.json
```diff
@@ -14,5 +14,24 @@
   "[json]": {
     "editor.defaultFormatter": "esbenp.prettier-vscode"
   },
-  "peacock.color": "#9b51e0"
+  "peacock.color": "#9b51e0",
+  "workbench.colorCustomizations": {
```
bgsandan, line 18: 😆

src/app/components/dialogs/sectionStartUpCostReviewDialog/sectionStartUpCostReviewDialog.component.scss
```diff
@@ -46,3 +46,9 @@ input {
   border: black 0.5px solid;
   text-align: center;
 }
+
+// user edited values
+input.ng-dirty {
+  color: $color_input_blue;
+  border-color: $color_input_blue;
+}
```
bgsandan, line 54: Nice! 👍

bgsandan (approved):
LGTM 👍

PR #1012, opened 2023-02-01
refactor: update angular to 15, update all dependencies, migrate angular material changes to new format

Fuhji:
- update angular to 15
- update all dependencies
- migrate angular material changes to new format
  - Spolier - there are a *lot* of changes for Material https://material.angular.io/guide/mdc-migration#1-update-to-angular-material-v15
- ci npm install ignore lock file to ensure getting dependencies from npm registry instead of nexus
- using fixed height in dialogs until can find 100% height workaround

# To test
- all the things! (actually, does it look like it did before, and if things arent obviously different that is good!)


Sveouu:
I like the new buttons. they look better in the first few pages.
there is some styling to be fixed in the CE section. 
- dialogs
- buttons placement  padding/margin

DanielLyleWilliamWarren (approved):

PR #1014, opened 2023-02-01
feat: div to display when api is not functional

jlczapski:
**Description**

Currently, if the API can't successfully return data there is no measure in place to warn users about this, they will just see a blank page.

**Solution**

Create a placeholder div that will notify users that there are issues with the API, along with a link to the status uptime robot page.

**To test**

Ensure that the div shows when the API is not operational.


Fuhji:
merge from develop and should help fix some of the checks above

Fuhji (approved):

PR #1017, opened 2023-02-06
Sr interventions

Sveouu:
- addded styling to input fields --> changing to blue if user entered
- added "close" button to baseDialog


Fuhji:
merge from develop and should help fix some of the checks above

src/app/components/dialogs/baseDialog/baseDialog.component.html
```diff
@@ -1,4 +1,5 @@
 <h1>{{ null != title ? title : '' }}</h1>
+<button class="close" mat-button (click)="onNoClick()">X</button>
```
DanielLyleWilliamWarren, line 2: What is the functionality of another close button?

Sveouu: I guess it is not really needed, the "back " button is sufficient, also the styling of the button is messed up with the new angular version.
Instead I feel it would be good that when clicking outside the dialog the cancel/back is triggered

DanielLyleWilliamWarren: if you set the 'closeable' in the dialog to true (in dialog.service.ts) when creating it the dialog will close when the user clicks outside of the dialog

Sveouu:
@DanielLyleWilliamWarren  removed additional close button... still need to resolve all those cypress failures

DanielLyleWilliamWarren (approved):

Fuhji (approved):

PR #1042, opened 2023-02-09
fix: various dialogs

### LINKED ISSUE #1029: Intervention Review General issues
- [x]  text wrap 

image

- [x]   Align table header cell with table data cell, looks like table headers are right aligned and table cells are left aligned. 

image

- [x]  'Add Micronutrient' button should only be visible on step 1 (Baseline Assumptions), remove from othes.

- [x]  GFDx buttons need to change to 'User Defined' if changes are made to the respective data.

- [x] padding on 'reset to default values' dialog needs updated.

- [x] Add generic '.bmgf-maps-nice-scrollbar' to y axis.

- [x] Resolve css issue where something is causing there to be an overflow on the x-axis of steps 2,3,4 & 5.

- [x] Intervention Selection dialog needs padding 

- [x] Add Mn name to header in image below: 
image

- [x] Make button position static on table dialogs.

- [x] Intervention list card Last Updated to use API returned time not current user's time

Fuhji:
- about ce dialog now loading (before the click event was on the icon not the button so never triggered)
- multiple dialogs now using standard layout rather than custom
- add intervention dialog content with extra spacing between elements
- dialog close button margins to avoid overlaps
- intervention description panel using live data instead of static
- fixed dialog action buttons to avoid scrolling to find the close
- close #1029

### DIFF
--- a/src/app/components/dialogs/baselineDescriptionDialog/baselineDescriptionDialog.component.html
+++ b/src/app/components/dialogs/baselineDescriptionDialog/baselineDescriptionDialog.component.html
@@ -1,7 +1,4 @@
-<app-base-dialog [data]="data">
-  <h2>Dietary supply of {{micronutrient}} in {{ country }}: Dataset Details</h2>
-  <hr>
-
+<app-base-dialog [data]="data" [title]="title">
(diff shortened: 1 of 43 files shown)


DanielLyleWilliamWarren (approved):
Tested and looking good!

PR #1046, opened 2023-02-10
feat: update add and load intervention dialog

Fuhji:
- fork/load intervention dialog updated to show latest interventions from api instead of only on first load
- load interventions input now a dropdown list of latest interventions instead of manual number input
- only interventions above ID 3 are allowed to be loaded. This is hard coded for now but can use intervention parameter when api enables it

## to test
- load an intervention from the dialog using the drop down list
- fork an intervention, then open the dialog again and confirm that the newly created intervention can be loaded from the dropdown list
- only interventions iD > 3 can be loaded from the dropdown list


jlczapski (approved):
All looks good!

Issue #1041, opened 2023-02-09
If no data is available then show 'Content Unavailable' message


DanielLyleWilliamWarren:
If a user selects a vitamin or mineral e.g Biotin, that does not have any data associated to it, then show a message instead of step 1 informing the user that intervention creation is not currently possible.

- [x] Display component informing user that no data available. - Input on content from @pautva / @bgsandan 
- [x] Disable 'Confirm and Continue' Button. 

Current screen:
![image](https://user-images.githubusercontent.com/38860838/218062922-2db0f0b5-af3f-4de4-b7bf-de3170e247f3.png)

PR #1047, opened 2023-02-10
Svea no mn

Sveouu:
closing: #1041 
- added NoDataComponent to a higher Level so that it can be used by multiple components.
- disabled buttons so users cannot proceed through interventions review if no MN data is available.


DanielLyleWilliamWarren (changes requested):
You'll have to delete your package.lock file and pull develop then push up again before we can merge

DanielLyleWilliamWarren (approved):

DanielLyleWilliamWarren (changes requested):
I'm just making a couple of changes locally!

DanielLyleWilliamWarren (approved):

PR #1048, opened 2023-02-10
feat: gfdx buttons now change when row edited in monitoring and industry

DanielLyleWilliamWarren:
- When user edits GFDx button on industry or monitoring info steps the button changes to 'User Inputed'


Sveouu:
Cannot view the deployed version, but the code looks good to me!

Fuhji (approved):
looks good - only comment would be to change 'user inputed' to either 'user input' or 'user inputted' as 'inputed' is a typo.

Sveouu (approved):

Issue #1030, opened 2023-02-08
Baseline Assumptions issues (step1): 


DanielLyleWilliamWarren:
- [x] Highlight input text to blue when user changes value.
- [x] Add padding on 'Add Micronutrient' dialog
- [x] Mn compound selection needs work. - Looks like same change event is being called when there are multiple added Mn compounds.
- [x] Reset to default data when user clicks or 'Reset to default values' button.
- [x] Resolve errors (images below). These are thrown when user clicks on 'Add micronutrient' button.
- [x] Re-add functionality to remove a micronutrient from premix
- [x] Ensure same micronutrient can't be added twice (?)
- [x] Selecting micronutrient in dialog causes duplicate values
- [x] Fix CSS issue in '?' dialog
- [x] PremixTable component is using hardcoded intervention id value.
- [x] Ensure changes to inputs are being cached once they are edited
- [x] Buttons should change from Gfdx to User Input if they have been edited (without reloading page)
- [x] Buttons should show User Input if they have been edited previously e.g `year0Edited = truefalse/null`
 
![image](https://user-images.githubusercontent.com/38860838/217533577-03417e39-8125-4bde-811c-3b91e82907ff.png)

![image](https://user-images.githubusercontent.com/38860838/217533594-98a89478-593a-45d3-b916-fa4a8f498ea3.png)

bgsandan:
- [ ] Display percentage values as 0-100 (still patch updates to the API as decimal).  (Row units are returned in the API response)
<img width="665" alt="image" src="https://user-images.githubusercontent.com/39411049/218710767-97c0d88f-03c8-4630-bffd-a55f55beee8a.png">
<img width="564" alt="image" src="https://user-images.githubusercontent.com/39411049/218711119-9be53056-85dd-4235-8188-f118bdbf555c.png">

- Round cell values to 2.dp (arguably this could be an API issue)

PR #1055, opened 2023-02-14
feat: baseline assumptions

jlczapski:
Almost all issues completed now, few things to check over but what has been done so far should be merged into develop so that people can pick up latest changes. 

@bgsandan added a few minor changes this morning to #1030 which I will resolve this afternoon.


src/app/components/dialogs/mnAdditionDialog/mnAdditionDialog.component.ts
```diff
@@ -68,4 +68,15 @@ export class MnAdditionDialogComponent {
   public handleSelectionChange(event: MatSelectChange): void {
     this.micronutrientsSelected = event.value;
   }
+
+  public alreadyAdded(micronutrientName: string): boolean {
```
Fuhji, line 72: can this be a Type instead of plain string? does that make it easier to check?

jlczapski:
Still to fix:

- [x] Duplicate remove buttons, mn is not actually removed
- [x] Reset to default button sets percentage values back to decimal
- [x] MN inputs don't change to 'user input' after updating
- [x] Fixed widths and paddings for MN

jlczapski:
To test:

- [ ] Can add & delete micronutrients into/from premix table
- [ ] Micronutrient inputs change to 'User input' when values are changed

DanielLyleWilliamWarren:
Should I be able to see the micronutrient in the list for example, Fe?

image

jlczapski:
> Should I be able to see the micronutrient in the list for example, Fe?
> 
> image

Good spot

DanielLyleWilliamWarren (approved):
Now, at the moment a user can't add multiple compounds of the same micronutrient. This is a bit of an issue that needs further investigating as a user should be able to select multiple compounds. I see you've disabled access for a user to add multiple of the same Mn atm.

PR #1056, opened 2023-02-14
style: update version dialog

Fuhji:
- updated version dialog to standardise with others

image

# to test
- from quick maps pages click the version number in footer


jlczapski:
image

Padding in between title/text looks a bit off, line height might need adjusting too. 

Is it obvious that the version number in footer can be clicked? I struggled to find the trigger at first.

@Fuhji

DanielLyleWilliamWarren (approved):

PR #1060, opened 2023-02-15
User edited inputs visible

Fuhji:
-  show user edited fields in industry and monitoring stages
  - [x]  compliance over time
  - [x] industry information
  - [x] monitoring information
  - [x] startup/scaleup
  - [x] recurring costs
  
image

- intervention object updated to include extra API params like last edited and parent ID
- intervention card now shows last edited date

todo: baseline page, but currently its being poked so steering clear for now

## to test

- [x] open intervention id 22 and click through pages, they should be highlighted where inputs have been edited
- [x] open another intervention id and be able to edit/highlight/PATCH/go back to confirm changes 
- [x] edit a field, confirm and continue, then go back and check that the edited field is still highlighted
- [x] if table has a source column (with user input/gdx), if says gdx, make a change, confirm that changes to 'user input', confirm and continue, then go back and confirm that the input is highlighted AND the source says user input
- [x] check intervention card 'last edited' date, go into intervention and make edit/save and close, then check the card date has been updated to now


Sveouu (approved):
works great!

PR #1066, opened 2023-02-16
Skeleton loaders for drop-in to anywhere with a table

Fuhji:
- using skeleton loader package that can drop into any html template
- added dynamic table skeleton template with custom rows and columns
- added skeleton loaders into tables within intervention pages for slow connections

## to test
- view each page of the intervention steps and watch for a skeleton table to show before the data loads and a table of content replaces it
- this can be helped by turning dev tools > network > throttling to something like Fast 3G, as a wired network will get the data too quick to notice the change


Sveouu (approved):
Very cool. Good to know about the throttling.

PR #1068, opened 2023-02-16
User inputs within intervention tables are disabled when the API specifies

Fuhji:
- User inputs within intervention tables are disabled when the API specifies

image

- updated the table legend to show the user edited as blue outline (instead of solid blue) and non editable input as grey

image

## to test
- click though each intervention page and if there are inputs, they should be disabled when the api specifies
- table legend is shown in the correct places alongside tables as well as in dialog tables

## notes

- #1069 highlighted by this PR but will be fixed when the dummy/test interventions are wiped


Sveouu (approved):
Looks good!

PR #1132, opened 2023-06-08
Fix Cypress tests 

jlczapski:
Some Cypress tests were failing since the URLs were missing parameter '&measure=food systems', so an error dialog was shown and hence Cypress couldn't run the UI tests properly.

There was also an issue with class selectors on `ui/quickMaps/baseline.ts`. A new boolean flag was recently added to toggle content in the HTML file `src/app/pages/quickMaps/components/estimate/estimate.component.html`, but the test was looking for selectors that did not exist in the new HTML content block. Updated the test to look for these new selectors.

### DIFF
--- a/cypress/integration/ui/quickMaps/baseline.ts
+++ b/cypress/integration/ui/quickMaps/baseline.ts
@@ -5,7 +5,7 @@ describe('Quick maps - Baseline', () => {
   });
 
   it('checks page for a11y compliance', () => {
-    cy.visit('/quick-maps/food-systems/baseline?country-id=MWI&mnd-id=Fe');
+    cy.visit('/quick-maps/food-systems/baseline?country-id=MWI&mnd-id=Fe&measure=food systems');
     cy.wait(8000);
     cy.injectAxe();
     cy.checkA11y(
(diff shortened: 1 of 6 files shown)


bgsandan:
:heart:

cypress/integration/ui/quickMaps/baseline.ts
```diff
@@ -37,14 +39,16 @@ describe('Quick maps - Baseline', () => {
...
-    cy.get('.vitamin > .estimateValue').should('be.visible');
-    cy.get('.vitamin > .estimateValue').should('have.text', '759');
-    cy.get('.baseline > .estimateValue').should('be.visible');
-    cy.get('.baseline > .estimateValue').should('have.text', ' -61% ');
+    cy.get('.median > .estimateValue').should('be.visible');
+    cy.get('.median > .estimateValue').should('have.text', ' 88.2 ');
+    cy.get('.adequacy > .estimateValue').should('be.visible');
+    cy.get('.adequacy > .estimateValue').should('have.text', '814');
+    cy.get('.prevalence > .wrap-15').should('be.visible');
+    cy.get('.prevalence > .wrap-15').should('have.text', '99.87% of sampled households (12396/12412)');
```
bgsandan, line 51: These tests are a tad flakey given the potential for data to change, but the edits here should "fix" the current test so that is good.

At some point we should refactor the test to locally call the API and then confirm the presented data matches the API response, but that can be for future Janusz :stuck\_out\_tongue\_closed\_eyes:

bgsandan, line 51: These tests are a tad flakey given the potential for data to change, but the edits here should "fix" the current test so that is good.

At some point we should refactor the test to locally call the API and then confirm the presented data matches the API response, but that can be for future Janusz :stuck\_out\_tongue\_closed\_eyes:

bgsandan (approved):
LGTM 👍

PR #1141, opened 2023-06-16
fix: interventions now loading after change to API#

DanielLyleWilliamWarren:
- Filters results on 'id' not 'name' that has been parsed.


bgsandan (approved):
LGTM :+1:

PR #1152, opened 2023-06-21
feat: JSONLogic implemented for industry and monitoring information intervention stages

Fuhji:
- Using JSONLogic lib to translate API formula which modified how form values are calculated
- Using centralised service for JSONLogic which is called by components which need a value calculating

## To test

- [ ] using ID 20, navigate to industry or monitoring information stages. Some fields are greyed out as they are not editable, these fields are the ones which have a formula applied to them
- [ ] change a value in the first row and see that the values in row 2 are updated based on the formula calculation


DanielLyleWilliamWarren (commented):
LGTM - tested using int 4

DanielLyleWilliamWarren (approved):

Fuhji:
Dev API endpoint is not returning the correct values nor updating when sending a PATCH. Setting this PR as draft until can confirm working as expected.

Fuhji:
> Dev API endpoint is not returning the correct values nor updating when sending a PATCH. Setting this PR as draft until can confirm working as expected.

The issue was that Intervention 4 was being used, which is a template and non-editable. Updated to use non template e.g. ID 20 and works as expected.

PR #1155, opened 2023-06-21
feat: updated interventionData service with functions to store recent…

### LINKED ISSUE #1145: Save recently viewed intervention IDs to local storage
As per #1144 loading a previously forked intervention will require entering/pasting an intervention ID into the dialog.

We would like to save the last _N_ interventions accessed on the current browser to local storage to allow prompting the user to resume a previously viewed intervention

*   [ ] Store IDs and name of accessed interventions to local storage in chronological order
*   [ ] Clear old IDs/names when a certain number (e.g. 5) of interventions have been stored

DanielLyleWilliamWarren:
- Store recently accessed interventions (max 5) in localstorage; 
- resolves #1145


Sveouu (approved):
Working as expected. solving issue https://github.com/micronutrientsupport/micronutrient-support-tool/issues/1145

PR #1156, opened 2023-06-22
refactor: ce form ux flow

jlczapski:
**Updates:** 

- Removed select boxes for steps 4, 5 & 6. These are automatically populated when a template is selected & the values are read-only, the data is still sent in the request. 
- Form is not shown until the user selects a template to fork from.
- Clearer messaging for when user selects a country that has no regions associated to it.
- 'Copy and edit' button is disabled until a name is given, country is selected & micronutrient is selected.

**To test:**

- [ ] Flow of creating an intervention still works as it did previously
- [ ] Country/region/micronutrient options not shown until user selects a template
- [ ] User cannot proceed until all required fields are filled out
- [ ] No errors occur during POST request when user hits 'Copy & edit'

### DIFF
--- a/src/app/components/dialogs/costEffectivenessSelectionDialog/costEffectivenessSelectionDialog.component.html
+++ b/src/app/components/dialogs/costEffectivenessSelectionDialog/costEffectivenessSelectionDialog.component.html
@@ -1,142 +1,119 @@
 <app-base-dialog [data]="dialogData" title="Select an intervention">
-  <div class="content-container">
-    <p>
-      Select one of the existing interventions and customise it for your scenario or load an existing
-      intervention by its ID.
-    </p>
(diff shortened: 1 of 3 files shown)


DanielLyleWilliamWarren (commented):
After creating an intervention in the new form it is no longer stored in the URL, so isn't visible on the interventions page after reviewing it. 

Load button is also disabled when selecting from intervention dropdown (current;y being refactored by Svea and myself so not as important)

jlczapski:
> After creating an intervention in the new form it is no longer stored in the URL, so isn't visible on the interventions page after reviewing it.
> 
> Load button is also disabled when selecting from intervention dropdown (current;y being refactored by Svea and myself so not as important)

Yeah this is being resolved in a separate branch, kept this one purely for the UX changes to the form.

Fuhji (approved):

PR #1157, opened 2023-06-22
feat: quickchart create short url via POST

jlczapski:
This [Postman](https://www.postman.com/speeding-flare-926667/workspace/quickchart-public-workspace/example/11849833-9aff1606-4b22-4ad1-99cf-81ce3a48ac2c) instance shows how to create a shorter URL via a POST request with Quickchart.

This works as expected and resolves the issue of the URL being rejected for being too long. However, I have noticed a handful of caveats related to using this approach:

- Doesn't seem to be any way to specify the returned chart to be PNG or PDF, it just returns a WEBP image.
- Our custom MAPS watermark is not added to the image.
- There is an error when trying to download an image of a chart that has type `treemap`.

Regarding these current issues, leaving this PR as draft for now until we have found out how to resolve them or accept the caveats with the POST approach.

### DIFF
--- a/src/app/services/quickChart.service.ts
+++ b/src/app/services/quickChart.service.ts
@@ -1,10 +1,14 @@
 import { Injectable } from '@angular/core';
 import { ChartJSObject } from '../apiAndObjects/objects/misc/chartjsObject';
+import { HttpClient } from '@angular/common/http';
+import { Observable } from 'rxjs';
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const QuickChart = require('quickchart-js');
 
 @Injectable()
 export class QuickchartService {
+  constructor(private http: HttpClient) {}
+
   /**
    *
    * @param chartData - object containing type, data, labels and options for your chart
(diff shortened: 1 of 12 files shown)


src/app/services/quickChart.service.ts
```diff
@@ -22,4 +26,15 @@ export class QuickchartService {
...
+
+  public getChartViaPost(chartData: ChartJSObject, format: string): Observable<unknown> {
+    const body = {
+      backgroundColor: '#fff',
+      width: 1000,
+      height: 500,
+      format: format,
+      chart: JSON.stringify(chartData),
+    };
+    return this.http.post('https://quickchart.io/chart/create', body);
```
bgsandan, line 38: This will be the cause of the lack of custom branding.  Our QuickChart instance is at https://quickchart.micronutrient.support

bgsandan, line 38: This will be the cause of the lack of custom branding.  Our QuickChart instance is at https://quickchart.micronutrient.support

jlczapski:
Postponing until updates to our custom Chartjs setup have been completed. Need to set up new API route to create short URL but with MAPS branding.

jlczapski:
This has been resolved by removing the use of Quickcharts.

PR #1169, opened 2023-06-29
feat: show percentages between 0 and 100 but keep api comms 0 and 1

Fuhji:
- intervention forms which have a user editable percentage display the value from the api (which is between 0 and 1) as a human readable percentage between 0 and 100. 
- Input set to `number` which allows the up/down increment UI on the input itself
- If the user manually types in a number which is below 0 (e.g. -50) or above 100 (e.g. 999) the input will correct itself to the nearest valid value i.e. 0 or 100. A notification will display advising the user of this. It was felt that instant correction was preferable to highlighting the input in red or displaying some less-clear messages due to the number of inputs which the user could potentially edit.

### to test

- [ ] calcium/ethiopia, ID 47
- [ ] each intervention page which has inputs for percentages should only allow values between 0 and 1
- [ ] increment on input should only allow 0 to 100 values
- [ ] notification should alert user when invalid value is input and value auto-corrects to nearest valid value (0 or 100)
- [ ] edited user values should be correctly identified for PATCH to API
- [ ] user edited values (between 0 to 100) should be converted to (0.0 to 1.0) when sending to the API

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -14,5 +14,24 @@
   "[json]": {
     "editor.defaultFormatter": "esbenp.prettier-vscode"
   },
-  "peacock.color": "#9b51e0"
+  "peacock.color": "#9b51e0",
+  "workbench.colorCustomizations": {
+    "activityBar.activeBackground": "#b47ce8",
+    "activityBar.background": "#b47ce8",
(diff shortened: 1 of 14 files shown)


DanielLyleWilliamWarren (approved):
LGTM!

PR #1180, opened 2023-07-12
feat: implement next version of jsonlogic to intervention section with wider rollout

Fuhji:
- several pages of the intervention review stage now are able to update input forms based on a formula which is included in the API request. These formula broadly fit into three use cases
- 1. formula of cell A is based on cell B which is available in the current API response
- 2. formula of cell A is based on a formula in cell B which is part is not available in the current API response. This uses an API caluclated value
- 3. formula of cell is based on values which are entirely in an external API response

## testing
- load Ethiopia/Calcium
- load intervention 46
- navigate through the sections and pay attention to SUSU costs and Recurring costs section.
- open some modals and edit the data, formula should update in dark/non editable cells

> Note: intervention section values are not being updated based on the user-edited values, so when loading a form the values shown are the 'default'. Editing a cell then triggers the recalculation which is why editing row 1 may update multiple other rows as they are *now* recalculating based on their formula inputs. When the API stores the new values this issue will be gone.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -14,5 +14,24 @@
   "[json]": {
     "editor.defaultFormatter": "esbenp.prettier-vscode"
   },
-  "peacock.color": "#9b51e0"
+  "peacock.color": "#9b51e0",
+  "workbench.colorCustomizations": {
+    "activityBar.activeBackground": "#b47ce8",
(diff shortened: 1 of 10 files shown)


DanielLyleWilliamWarren:
LGTM

DanielLyleWilliamWarren (approved):

PR #1183, opened 2023-07-13
build: update es version and update common js deps

Fuhji:
- making the build cleaner (fixing warnings) by updating some build settings to match current Angular version
- update to ES2022 build
- update commonjs libs
## to test
- [ ] Business As Usual functionality i.e. nothing should have changed


DanielLyleWilliamWarren (approved):
Explored application and no issues

PR #1184, opened 2023-07-14
fix: ui elements in the intervention stages

Fuhji:
- submarine text within header cells fixed
- where appropriate, vertical align header text as less jarring than central alignment
- startup/scale up costs section media query update to fix overlapping content
- left-aligned header to match rest of table ui
### to test

- [ ] baseline/comliance pages, clicking header ? button opens dialog and does not move the text content down
- [ ] startup/scale up, shrink browser and divider hides before hits left text, responsive row to column happens sooner o avoid overlap


DanielLyleWilliamWarren (approved):
Tested, LGTM!

PR #1191, opened 2023-07-14
feat: integrate authentication with API calls

### LINKED ISSUE #1177: Integrate authentication with intervention loading/creation
The routes for creating and loading interventions now optionally support passing of `X-Session-Token` to provide additional functionality for logged in users:

*   GET `/interventions` - Populates the dropdown list of template interventions when creating an intervention
    *   When not logged in - only returns system template interventions to be forked
    *   When logged in - returns system template interventions and interventions belonging to the logged in user
*   GET `/interventions/<id>` - Fetches the details about a single intervention for loafing an existing intervention
    *   If the intervention is a template, nothing is returned (templates shouldn't be loaded directly, they should be forked)
    *   If the intervention has no registered owner (Anonymous) it is always loaded
    *   If the intervention has a registered owner, a session token for that user must be provided to display the data
*   POST `/interventions` - If a session token is provided, the created intervention is assigned to the logged in user, otherwise it is anonymous

Currently the data loading endpoints for the various pages do not respect the session-token (i.e. an anonymous request to industry-information for an intervention belonging to a registered user would still return data), though the UI using the routes above shouldn't allow this to occur.  These routes should still have the session-token appended and their functionality will be appropriately locked down in the near future

There is also a new route:

*   PATCH `/interventions/<id>` - _Requires_ a session token.  Allows a logged in user to 'claim' an anonymous intervention which then gets added to their account

DanielLyleWilliamWarren:
- Integrates authentication into GET `intervention` (dictionary), GET `/interventions/<id>` , POST `/interventions` and a new call PATCH `/interventions/<id>` 

- resolves #1177


bgsandan:
LGTM 🚀

bgsandan (approved):

PR #1197, opened 2023-07-20
feat: update cost effectiveness route

Sveouu:
- [x] resolves issue 1172
- [x] cost effectivess removed from quicklmaps and  routing established
- [x] reformatted intevrntion ID copy/email section

To test:

1. Navigate to Cost Effectiveness, via Quickmaps OR map tools.
2. Create an intervention / load a recently created one
3. Go through form, all pages should load / function.
4.( Cost Summary is failing to load data, this is an issue prevalent in the develop branch too )


Fuhji:
image


image


- Issue 1: Chrome - The left sidebar no longer shows up when accessing Quick Maps after the map selection stage. I have cleared cache etc and still get the same.
- Issue 2: Firefox - sidebar appears more regularly, but, when re-visiting the map country selection page as soon as you click a country it navigates away to the next page rather than waiting for the user to confirm their selection of country and micronutrient

Fuhji:
Update: issue 1: missing sidebar no longer seems to be an issue

Issue 2: 

- clear cache etc
- home > quick maps
- select country (view results button disabled)
- select mineral (view results button enabled)
- click 'view results'
- navigate to home > quick maps
- change country, click on map or via list, navigation happens straight away instead of waiting for 'view results'

Might not be a big issue but the UX is different from the first time you access the page to the subsequent ones

jlczapski (approved):

PR #1203, opened 2023-07-26
feat: intervention storage update

DanielLyleWilliamWarren:
- Creates a better way of tracking interventions created rather than prompting a user to enter an id of one they've already created.
- Various styling changes

To test:

- [ ] log in
- [ ] load several of your own interventions.
- [ ] find an anonymous one (48 is currently anon), API errors atm when using copy and edit functionality.
- [ ] refresh the page, they should be remain.
- [ ] logout, only the anon interventions should remain..
- [ ] try to claim, should prompt login.
- [ ] login and claim intervention, intervention should update and claim functionality disappear.


Fuhji (approved):
LGTM 👍 just need to rename the PR title to meet commitlint guides and ready to go

PR #1212, opened 2023-07-28
feat: replace fab menu package

DanielLyleWilliamWarren:
- Replaces old `@angular-material-extensions/fab-menu": "^6.1.0"` dependency with native code to remove ng16 upgrade blocker 

To Test:

- [ ] Check fab menus on quickmaps pages


bgsandan (commented):
Functionality looks good, just looks like there are some cypress a11y issues, and the PR name to update :slightly\_smiling\_face:

PR #1316, opened 2024-01-12
feat: update biomarker obj, add skellies, update sidenav, update params

DanielLyleWilliamWarren:
- Creates Biomarker object and child objects
-  Adds skellies to display when data loading
-  Update side-nav to search specifically for Biomarker data
-  Prune biomarker component
- Update biomarker status and table to display Biomarker aggregation threshold data
- Update url params to store biomarker query params and search appropriately on load


Fuhji (approved):
LGTM 👍

PR #1317, opened 2024-01-16
feat: add objects for extra biomarker response fields.  Basic wiring and display

bgsandan:
Add API objects for the new binnedData and totalThresholds response objects

Basic wiring to display total dataset thresholds in header bar (replacing hardcoded 22% values!)

image

Basic wiring of the binnedData response to the histogram in the bottom card


Fuhji (approved):
LGTM

PR #1318, opened 2024-01-17
fix: boxplot chart for biomarker data updated for new api data structure

Fuhji:
tested using Ethiopia / Iron


bgsandan:
The boxpolots look good.  Would it be possible for them to move up to the "Chart" tab in the upper card (as this is where the data relating to the aggregations comes from).  Also can we have the histogram back in the bottom chart? 🙃 

Sorry if it wasn't clear where the different charts were supposed to be 😢

bgsandan:
LGTM 🚀

bgsandan (approved):

